PR #2, opened 2015-06-30
Cdn

PeterDaveHello:
To speed up and make the repo thinner, easier to manage.


jackycute:
Hello, very good to see you done this. I agreed with your point.
But if someone want to use this locally then will be a problem using cdn.
Is there any other choice can make?

PeterDaveHello:
But you are already using cdn, e.g. 
https://github.com/jackycute/HackMD/blob/master/public/index.html#L18

So I don't think this is a new issue.

Produce a local version may be a possible solution.

jackycute:
Oh my, your totally right, I missed that...
OK, I will merge your PR, thank you very much.

I'm consider using requireJS to do cdn fallback now.

PeterDaveHello:
:+1:

jackycute:
Hello, I might fallback and not using cdn.
Because I found many people like to use this offline and internally, so it's better not use cdn.
Thank you anyway, andwe might have a plan to make a standalone app in the future.

PeterDaveHello:
Maybe we can use a config to decide use it or use local files?

jackycute:
That's a good idea, I will try it.

PeterDaveHello:
I think config is good enough :smile:

jackycute:
Hey, I've add useCDN option, take a look!

PeterDaveHello:
:+1:

PR #20, opened 2015-09-22
bugfix: useCDN option must be passed to ejs_renderer

193s:
https://github.com/hackmdio/hackmd/commit/b6c758f2fc2cb502ed194f2700e033564f2521f4

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -50,7 +50,8 @@ function responseError(res, code, detail, msg) {
         filename: template,
         code: code,
         detail: detail,
-        msg: msg
+        msg: msg,
+        useCDN: config.usecdn
     });
     res.write(content);
     res.end();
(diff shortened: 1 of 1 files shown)


lib/response.js
```diff
@@ -92,11 +93,13 @@ function responseHackMD(res, noteId) {
         var template = config.hackmdpath;
         var options = {
             cache: !config.debug,
-            filename: template
+            filename: template,
+            useCDN: config.usecdn,
```
jackycute: No need to pass useCDN here.

jackycute:
Oh yes, I forgot these parameters should pass to ejs, it's my fault.
But there is a redundant passing there, would you fix that? And I will merge this.

193s:
fix & rebased. thanks!

PR #46, opened 2015-12-09
Update package.json

PeterDaveHello:
Change required dependency - express from "4.x" to ">=4.5"

Reference(security issue):

https://nodesecurity.io/advisories/express_no-charset-in-content-type-header


jackycute:
Thanks a lot
Security is very important here.

PR #101, opened 2016-04-23
Hide unused UI elements

janniklorenz:
First of all, 0.4 is a great release, and the single database dependency now simplifies everything a lot.

2 Points I fixed:
- Login button and message on history page only shows when we have a login method enabled
- Dropbox and Google Drive export only available, when configured

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,8 @@ Third-party integration api key settings
 | facebook, twitter, github, dropbox | `config.json` | for signin |
 | imgur | `config.json` | for image upload |
 | dropbox | `public/views/foot.ejs` | for chooser and saver |
-| google drive | `public/js/common.js` | for export and import |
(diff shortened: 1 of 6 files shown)


jackycute:
Hi, Lorenz!
Thanks a lot for contributing these commits.

But I saw some duplicated statement in 
https://github.com/hackmdio/hackmd/pull/101/files#diff-7df48c29918a6d59c13923903e548cdaR36
https://github.com/hackmdio/hackmd/pull/101/files#diff-7df48c29918a6d59c13923903e548cdaR131
Could you fix that? I would like to merge this PR :smile:

janniklorenz:
Sure, should be working now.

jackycute:
Hi, Lorenz
I feel like to merge this PR.
But I think of that the google Drive and dropbox drop-in are client-side features.
I'll need do some other commits after merge this PR.
So give me few days.
Thanks still.

PR #102, opened 2016-04-23
Fix Esc key in Vim mode

Zankio:
Esc key should work in the vim mode

### DIFF
--- a/public/js/index.js
+++ b/public/js/index.js
@@ -5,13 +5,14 @@ var defaultExtraKeys = {
       cm.setOption("fullScreen", !cm.getOption("fullScreen"));
     },
     "Esc": function(cm) {
-      if (cm.getOption("fullScreen")) cm.setOption("fullScreen", false);
+       if (cm.getOption('keyMap').substr(0, 3) === 'vim') return CodeMirror.Pass;
+       else if(cm.getOption("fullScreen")) cm.setOption("fullScreen", false);
     },
     "Cmd-S": function () {
-        return CodeMirror.PASS
+        return CodeMirror.Pass
     },
     "Ctrl-S": function () {
-        return CodeMirror.PASS
+        return CodeMirror.Pass
     },
     "Enter": "newlineAndIndentContinueMarkdownList",
     "Tab": function(cm) {


jackycute:
Thank you Zankio!
But I leave some comments on your commit.
Please follow it and do some fix and I will merge it, ok?

public/js/index.js
```diff
@@ -5,13 +5,14 @@ var defaultExtraKeys = {
       cm.setOption("fullScreen", !cm.getOption("fullScreen"));
     },
     "Esc": function(cm) {
-      if (cm.getOption("fullScreen")) cm.setOption("fullScreen", false);
+       if (cm.getOption('keyMap').substr(0, 3) === 'vim') return CodeMirror.Pass;
```
jackycute: I think you can do a complete compare here since no other input mode is similar to `vim` like `cm.getOption('keyMap') === 'vim'`

public/js/index.js
```diff
@@ -5,13 +5,14 @@ var defaultExtraKeys = {
...
+       else if(cm.getOption("fullScreen")) cm.setOption("fullScreen", false);
     },
     "Cmd-S": function () {
-        return CodeMirror.PASS
+        return CodeMirror.Pass
```
jackycute: please use `PASS` instead of `Pass` for consistentency

public/js/index.js
Zankio: `cm.getOption('keyMap')` will return `vim-insert` In the insert mode;

image

public/js/index.js
Zankio: If you want to use default behavior, `PASS` is not working.
`CodeMirror.PASS` return `undefined`
`CodeMirror.Pass` return `Object { }`

[CodeMirror User Manual](https://codemirror.net/doc/manual.html#keymaps)

>  A key handler function may return CodeMirror.Pass to indicate that it has decided not to handle the key, and other handlers (or the default behavior) should be given a turn.

Or I got it wrong.

jackycute: Oh, you're right!
Then I have to change all my `CodeMirror.Pass` usage.
Sorry I must followed so old standard or some bad snippet.

public/js/index.js
jackycute: Sorry about my ignorance.
I'm not using vim keyMay very much.

PR #121, opened 2016-05-12
Export/Import Gitlab Snippets

jccrofty30:
This PR extends to allow for exporting/importing connected GitLab account snippets.


jccrofty30:
Actually think I included the GitLab authentication updates in here as well. Would still pull that in separately first.

jackycute:
Oh, dude. You're awesome!
Wait until weekends, I will merge all branches 👍

jccrofty30:
Change your mind about weekend rollin?

jackycute:
Nope, WIP now.

jackycute:
Oh, I found that you add new columns to db.
We need a db migration script now.
See more here: https://github.com/sequelize/express-example

jackycute:
I will add the migrations, just wait.

jackycute:
Hey, I merge all the PRs.
But you need to do some migrations, see more in README.
Thanks for providing such great PRs.

jccrofty30:
Sorry I was AFK, will review everything.

app.js
```diff
@@ -421,6 +442,29 @@ app.post('/uploadimage', function (req, res) {
         }
     });
 });
+//get gitlab parameters
+app.get('/gitlab', function (req, res) {
```
jackycute: Here, I don't think use a root path `/gitlab` is a great idea.
Don't easily use root level path, because it will consume other path usage.
Since gitlab have a auth path, maybe we should move this action to there,

app.js
```diff
@@ -421,6 +442,29 @@ app.post('/uploadimage', function (req, res) {
...
+    var ret = { baseURL: config.gitlab.baseURL };
+    models.User.findById(req.cookies.userid)
```
jackycute: Please look at `/me` api, it already show how we handle to find a user, no need use cookie parser again.

PR #126, opened 2016-05-23
Dump all notes to a git repo for backup purpose

DrMurx:


bin/dump-to-git.js
```diff
@@ -0,0 +1,131 @@
...
+
+var fs = require('fs');
+var process = require('process');
+var child_process = require('child_process');
+var Promise = require('bluebird');
+var NodeGit = require('nodegit');
+var LZString = require('lz-string');
+var models = require("./lib/models");
+
+var repoRootDir = './notes';
```
ccoenen: this should be configurable, at some future point.

bin/dump-to-git.js
```diff
@@ -0,0 +1,131 @@
...
+    return repo.getCommit(headRef);
+  });
+}
+
+
+function commit(repo, tree) {
+  return getHeadCommit(repo)
+  .then(function (head) {
+    console.log('Commiting...');
+    var sig = NodeGit.Signature.now('Committer', 'johndoe@localhost');
```
ccoenen: can nodegit take the regular configuration from git itself? (`git config user.name` / `git config user.email`)
If it can, i'd leave this part out.
If it can't this should definitely be configurable.

bin/dump-to-git.js
```diff
@@ -0,0 +1,131 @@
...
+    return repo.createCommit('HEAD', sig, sig, 'Automated import', tree, [head]);
+  });
+}
+
+
+
+NodeGit.Repository
```
ccoenen: this whole following structure is very deeply nested promise code. Currently, I find this rather hard to read. I think 2-3 things should be broken out as regular named functions, so that the overall program flow is easier to follow.

bin/dump-to-git.js
```diff
@@ -0,0 +1,131 @@
...
+process.chdir(repoRootDir);
+
+
+function exportNotes() {
+  return models.Note
+  .findAll()
+  .then(function (notes) {
+    // Iterate over all notes
+    return Promise.all(notes.map(function (note) {
+      var file = LZString.compressToBase64(note.id);
```
ccoenen: one could also try to parse the optional yaml metadata at the top of a note. Or use the first headline.

Only downside is, that they may not be unique. But you'd get readable filenames in return.

bin/dump-to-git.js
```diff
@@ -0,0 +1,131 @@
...
+      .then(function (tree) {
+        return commit(repo, tree);
+      });
+    })
+    .then(function () {
+      console.log('Get remote...');
+      return repo.getRemote('origin')
+    })
+    .then(function (remote) {
+      console.log('Push to origin');
```
ccoenen: to push or not to push is also a candidate on my "to configure" list. Maybe this could be dependent on the existence of a specific remote. Something along the lines of "if `git remote hackmd-backup` exists, push to it, no questions asked. But that's very implicit.

bin/dump-to-git.js
```diff
@@ -0,0 +1,131 @@
+'use strict';
+
+var fs = require('fs');
+var process = require('process');
+var child_process = require('child_process');
+var Promise = require('bluebird');
+var NodeGit = require('nodegit');
+var LZString = require('lz-string');
+var models = require("./lib/models");
```
ccoenen: This yields a "Error: Cannot find module", when the script lives in `bin/` (where it's supposed to be, judging from this pull request). I have to move the script to `/` to run it beyond this point.

bin/dump-to-git.js
ccoenen: Also, this directory has to be pre-existing and be a git repo, otherwise you run into 
`ENOENT: no such file or directory` or `Could not find repository from '.'`

I think the script should check for that and create the directory and init a git repo.

bin/dump-to-git.js
ccoenen: Also, adding ".md" to the filename would be a nice addition. That way things like gitlab would render it as markdown as well.

bin/dump-to-git.js
```diff
@@ -0,0 +1,131 @@
...
+
+
+NodeGit.Repository
+.open('.')
+.then(function (repo) {
+
+  return exportNotes()
+  .then(function(files) {
+
+    return getHeadCommit(repo)
```
ccoenen: I believe this fails for completely empty repos as well. I can't be completely sure, but on the very first run (with a git repo with no commits), I get this error:

`Reference 'refs/heads/master' not found`

And the script does not create a commit for me, then. Export to files works, though.

bin/dump-to-git.js
jackycute: Because these packages should add to `packages.json` dependency.
And also need the run `npm install` (or `bash bin/setup`) to install them.

jackycute: oh sorry I missed it, that path is wrong (obviously).
it should be `../lib/models`.

DrMurx: Yeah, sorry. Was a bit sloppy here when I hacked that script. I'll update.. 😏

bin/dump-to-git.js
DrMurx: True. Nobody claimed that you can't end in callback hell even with promises...

ccoenen:
I changed three things that were most important to me:

- the model path
- config via `config.json`
- filename in the target directory

@DrMurx I created a branch of my own, in case you'd like to add them to the PR:
https://github.com/ccoenen/hackmd/tree/git-export

rufuspollock:
Love to see this feature in!

ccoenen:
@rufuspollock  for now, you can just take the file and use it. The only thing that's conflicting is the package.json, which is to be expected as we add bluebird and nodegit. If you `npm install` them manually, you can just take the .js file and run it. That's what I do on my server right now.

```
NODE_ENV=production
10 4 * * * bash -lc 'cd /var/www/hackmd; /usr/bin/node bin/dump-to-git.js'
```

DrMurx:
I'll close this PR, it will be continued in #303.

PR #158, opened 2016-08-19
Added support for french localization (fr-FR).

Lapin0t:
Hey, thanks @jackycute for i18n support! Here is the french localization. I did it quite fast so everything may not be optimal (especially things like _revision_ or _snippets_ which don't really have french equivalents) but I double checked and at least there should be no typo/spelling error.


jackycute:
Thanks @Lapin0t

PR #168, opened 2016-09-06
Removed redundant condition.

zored:


jackycute:
So true, how did I ever commit these lol.

PR #169, opened 2016-09-06
Improved some translations

svg153:
I have translated the strings that were in English and improved some translations by the application context.


jackycute:
Thank you @svg153 !

PR #172, opened 2016-09-26
Remove expiry from cookies

j6k4m8:
As per [this issue](https://github.com/expressjs/session/issues/365)


j6k4m8:
I experienced what appears to be an isolated case in which, using the dockerized version of this repository, my cookies got out of sync with what the server was expecting, and the cookie expiry was _before_ the current time. This resulted in a `AUTH failed: Cookie is invalid.` console error, and all pages but the homepage (`/`) failed to load (eternal loading-screen).

By removing this line, we can mitigate this issue.

Issue #157, opened 2016-08-19
Need help for translation


jackycute:
Since we support i18n, do anyone could help to translate the text to their language?
For the English texts, please refer to https://github.com/hackmdio/hackmd/blob/master/locales/en.json

And please tell me what's the language you using and what's your culture name.
See more here https://msdn.microsoft.com/en-us/library/ee825488(v=cs.20).aspx

Lapin0t:
Hey, I just made pull request #158 to support fr-FR.

jackycute:
@a-r-g-v @193s Hello guys!
Would you like to help doing Japanese translation?
Much appreciated :smile:

jackycute:
Could we have Korean translation?
If we have it, then CJK will all coverd.

ivanorsolic:
Hey, I've just made a pull request (#193) for the Croatian translation. The language culture name for Croatian is hr-HR.

bSzala:
I will add Polish translation

Kietzmann:
Added Ukrainian translation
[link](https://github.com/hackmdio/hackmd/pull/197)

paraschadha2052:
I will add Hindi translation.

wolframtheta:
Can I translate it to Catalan?

jackycute:
@wolframtheta Sure, just make a PR.

wolframtheta:
@jackycute Do you grant me access to that repo?

jackycute:
Actually, you need to fork this repo to you and send Pull Request back here.
The english text strings right here: https://github.com/hackmdio/hackmd/blob/master/locales/en.json
Please modify above file from your fork and make a PR and I will merge it.
Thanks!

rafajaques:
Do you already have French? I can take this one.

jackycute:
Hi @rafajaques !

We already have French (fr-FR) from @Lapin0t,
https://github.com/hackmdio/hackmd/blob/master/locales/fr.json
You could review it and make it better or if you have another French culture variety will also be taken.

Thanks for your attention!

PR #178, opened 2016-10-01
add greek locales

zachariast:
Greek locales
More infos: el-GR	Greek - Greece	0x0408	ELL


jackycute:
Thanks @zachariast !

jackycute:
refer #157

PR #187, opened 2016-10-04
Fix README markdown

Yukaii:
### Changes

* add extra line after heading([MD022](https://github.com/DavidAnson/markdownlint/blob/v0.2.0/doc/Rules.md#md022---headers-should-be-surrounded-by-blank-lines))
* trim trailing whitespace
* change tab(hard tab) in [Structure section](https://github.com/hackmdio/hackmd#structure) to white spaces, make it aligned beautiful again 😆 
* add link for _operational tranformation_

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,25 +3,28 @@ HackMD
 
 [![Join the chat at https://gitter.im/hackmdio/hackmd](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/hackmdio/hackmd?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 
-HackMD lets you create realtime collaborative markdown notes on all platforms.  
-Inspired by Hackpad, with more focus on speed and flexibility.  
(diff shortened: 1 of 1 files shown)


jackycute:
Thank you @Yukaii !
First seeing that markdownlint stuff, really helpful.

jackycute:
I've also noticed other rules in the lint.
Will try to fit them into current editor behaviors.

jackycute:
As you see, when you trim the trailing spaces for new line, it will break the layout on the GitHub.
HackMD use breaks while GitHub use 2 trailing spaces for break.
But the README.md here is more for GitHub, may consider add them back.
What do you think?

Yukaii:
screen shot 2016-10-04 at 5 30 52 pm

I haven't noticed yet, sorry about that.

In markdown, we might need an extra line to make correct line break 😓

jackycute:
So how about we add back the trailing spaces, it's necessary for GitHub.
The lint rules also mentioned that if you use two spaces for new line, it's allow to change that rule.

Yukaii:
OK!

PR #197, opened 2016-10-10
Created Ukrainian Localization

Kietzmann:


jackycute:
Thanks @Kietzmann !

Please notice the country of Ukrainian is `uk` instead of `ua`.
I will change that for you.

PR #195, opened 2016-10-09
Use Webpack to bundle frontend code

Yukaii:
### Changes

* Bundle frontend code using webpack 🎉 
* Change main package manager to npm
* You can write `module.exports` now 👍 
* Load `viz.js` and render asyncronously through webpack chunk 💯 

### TODOs:

- [x] make sure public note page works
- [x] move some vendor javascript back to bower
- [x] fix slider mode
- [x] fix index js
- [x] fix jquery dependencies
- [x] re-check every packages I've changes
- [x] webpack production config
- [x] merge with latest master
- [x] fix slide view 
- [x] assets hash
- [x] check cdn setting

### Plan in future PR
- [ ] ~~We love es6~~

PR #202, opened 2016-10-13
Webpack Recovery

Yukaii:
#195 is a big change, which comes with webpack and module support. However, it also breaks a lot 😓 . In this PR, I will recover those broken functionalities.

### Fixes

- [x] Updated/Created timestamps
- [x] syncscroll link/unlink button
- [x] scrollSpy
- [x] cheatsheet is missing
- [x] history delete functionality is broken
- [x] HTML exporting
- [x] pageInit() not called in cover.js
- [x] spellcheck is broken
- [x] Can't change spaces width


jackycute:
Awesome!

PR #212, opened 2016-10-16
Sample editorconfig & eslint

Yukaii:
Prevent our favorite editor from accidentally trimming trailing whitespace 😓 .

### Changes

* Create `.editorconfig`. Current indentation is using space and width 4.
* Generate eslint config


jackycute:
Lint them all!

PR #230, opened 2016-10-23
Use CDNJS by default with https and SRI support

PeterDaveHello:
As I discussed with @jackycute earlier, this PR will

1. force enable https rather than just use protocol relative url for the security and http/2.0(and spdy) feature
2. try to use single cdn, currently CDNJS to maximize the benefits from http/2.0(and spdy) - multiplex the connections
3. enable SRI - subresource integrity - for security enhancement


jackycute:
Thanks a lot @PeterDaveHello 👍

PR #243, opened 2016-11-03
Google font optional cdn

Yukaii:
### Changes

* Add Source Sans/Serif/Code fonts to repo
* Config `useCDN` option for these fonts

### Font format conversion

The fonts downloaded from Google were converted by [webify](https://github.com/ananthakumaran/webify/) tool. The original file format is `ttf`.


jackycute:
Thanks!

PR #248, opened 2016-11-14
Support other options for image uploading

Yukaii:
### TODOs

- [x] local filesystem
- [x] s3
- [x] enable s3 config in env
- [x] s3 guide: policy setting
- [x] image security - using [sharp](https://github.com/lovell/sharp) lib
    - [x] [heroku build config](http://sharp.readthedocs.io/en/stable/install/#heroku)
- [x] update README

### Screentcast

hackmd-filesystem-upload


Yukaii:
Implement s3 image uploading using AWS javascript SDK.

### References

* http://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/global-config-object.html
* https://devcenter.heroku.com/articles/s3-upload-node
* http://stackoverflow.com/questions/17930204/simple-file-upload-to-s3-using-aws-sdk-and-node-express

```javascript
var AWS = require('aws');
var config = new AWS.Config({
  accessKeyId: 'AKID', secretAccessKey: 'SECRET', region: 'us-west-2'
});

s3 = new AWS.S3(config);
var params = {
  Bucket: 's3-bucketname',
  Key: file.name,
  Body: data
};

s3.putObject(params, function (perr, pres) {
  if (perr) {
    console.log("Error uploading data: ", perr);
  } else {
    console.log("Successfully uploaded data to myBucket/myKey");
  }
});
```

jackycute:
@Yukaii Great PR! Thanks a lot!
I'm also considering the security issue beneath the images.
It's better to remove EXIF and process them before actually saving them.
And it's important to limit the support image formats and use some lib to help.
For example: https://github.com/lovell/sharp

😄

jackycute:
BTW, I think this should return full path with `config.serverurl` prefixed, or it might not able to access in some cases.

Yukaii:
@jackycute 

I'm not sure whether `sharp` will have huge impact on performance. 😓  We might need more optimization on this in the future.

jackycute:
I just looked at sharp's README.md.
And it show up that they process the image via pure lib so it will still remain non-IO-blocking.
Hope it won't impact the server performance, we'll see.

ccoenen:
I have to admit I did not look at the code before asking, but: could this extend to file upload beyond images? Sometimes this might come in handy to also store - i don't know - a `.doc` or `.zip` alongside your hackmd pad.

Yukaii:
@ccoenen In this PR, I only focus on image uploading & processing. IMO text-based editing is the main goal of HackMD, similar products like Hackpad or Quip have limited support for file uploading.

To implement this feature, we'll need to open another api for handling file uploading specifically(s3 or filesystem), checking file size(maybe unlimited for self-hosting version) and file type in frontend code. You could open another issue to track this. 😸

ccoenen:
No problem!

PR #295, opened 2016-12-22
Fix heroku config file error and set default value

bananaappletw:
Fix  heroku config file error and set default value.
Easier to deploy on heroku.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,11 @@ You can quickly setup a sample heroku hackmd application by clicking the button 
 
 [![Deploy](https://www.herokucdn.com/deploy/button.svg)](https://heroku.com/deploy)
 
+You have to setup the minimal configuration
+
+- App Name ( Ex: hackmd )
+- HMD_DOMAIN ( Ex: hackmd.herokuapp.com )
+
 [migration-to-0.4.0](https://github.com/hackmdio/migration-to-0.4.0)
 ---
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -22,6 +22,11 @@ You can quickly setup a sample heroku hackmd application by clicking the button
 
 [![Deploy](https://www.herokucdn.com/deploy/button.svg)](https://heroku.com/deploy)
 
+You have to setup the minimal configuration
+
+- App Name ( Ex: hackmd )
+- HMD_DOMAIN ( Ex: hackmd.herokuapp.com )
```
Yukaii: I do prefer to let Heroku handle domain itself. I'm not setting `HMD_DOMAIN` value, and both heroku domain(xxx.herokuapp.com) and my custom domain work. 😄

app.json
```diff
@@ -36,17 +36,19 @@
             "required": false
         },
         "HMD_ALLOW_ORIGIN": {
```
jackycute: So here actually a mistaken from @Yukaii 
It should be `HMD_PORT` instead of `HMD_ALLOW_ORIGIN `.

app.json
```diff
@@ -36,17 +36,19 @@
...
-            "description": "web app port",
+            "description": "domain to allow cross origin request",
```
jackycute: don't change this

app.json
```diff
@@ -36,17 +36,19 @@
...
             "required": false,
-            "value": "80"
+            "value": "localhost"
```
jackycute: don't change this

bananaappletw:
Sorry, I misunderstood the meaning of config file.

PR #320, opened 2017-01-18
Add ability to configure API request scope. Fixes #301

### LINKED ISSUE #301: GitLab OAuth requests too much permissions
When logging in to HackMD via GitLab, the latter tells me the requested scope is too broad, as I am an admin. How could one request a smaller scope, which only receives what is needed from a user?

SISheogorath:
See https://github.com/fh1ch/passport-gitlab2/blob/master/lib/strategy.js#L46

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,7 @@ Environment variables (will overwrite other server configs)
 | HMD_GITLAB_BASEURL | no example | GitLab authentication endpoint, set to use other endpoint than GitLab.com (optional) |
 | HMD_GITLAB_CLIENTID | no example | GitLab API client id |
 | HMD_GITLAB_CLIENTSECRET | no example | GitLab API client secret |
+| HMD_GITLAB_SCOPE | `read_user` | GitLab API requested scope |
 | HMD_DROPBOX_CLIENTID | no example | Dropbox API client id |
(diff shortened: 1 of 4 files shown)


SISheogorath:
Please notice: untested

SISheogorath:
This should not effect any existing setup. As it only changes new config files generated from the `config.json.example` file.

As the default setting of the upstream plugin is to use `api` all already existing configs will continue as they already work. But improves security for new configs as it limits the authentication to the new `read_user` scope added in [8.15](https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/5951)

jackycute:
Hi @SISheogorath ,
Thanks for helping this out.
But I don't know if this is ok because we have ability to import/export snippet to gitlab and that might need api scope permission.

jackycute:
I think we could add option for gitlab snippet import/export feature.
And we could set the scope based on that option.

jackycute:
[Fix gitlab request permission by config](https://trello.com/c/5u6ay7od/6-fix-gitlab-request-permission-by-config)

jackycute:
Thanks @SISheogorath 
Fixed in 506a381eca5e2bdc31d3529ecd3f1f7797551406

PR #321, opened 2017-01-20
Handle SIGTERM the same way SIGINT is handled

DrMurx:
SIGTERM is now handled properly and performs a graceful shutdown the same way SIGINT does.

Before, SIGTERM couldn't kill the process, which eventually caused the operating system or dockerd to issue a SIGKILL to terminate the server unconditionally.


jackycute:
Thanks @DrMurx 
This looks great!

PR #325, opened 2017-01-21
Add basics for secret management by Docker 1.13

SISheogorath:
See https://docs.docker.com/engine/swarm/secrets/

This is a first draft (That's why not all secrets are covers right now). I'm open for suggestions. Feel free to comment :)


SISheogorath:
@jackycute can you please take a first view? Just to make sure its not work that would find its way into the project.

As there are some very special "*hacks*" in it ;)

jackycute:
It looks not bad.
Just few code styles need to be corrected.
Also I might need some docker users to test or tell how this needed.

SISheogorath:
Should be more or less finished here. Maybe docs in the README are missing

PR #326, opened 2017-01-21
ES6 integration

Yukaii:
### Notable Changes

- Rename npm script (Revert ed83dfc for convention)
- Add babel, transform most frontend code to es6 style using [lebab](https://lebab.io)
- Remove `config.js` in Frontend. Since we already have a `config.json` in project root, just use it. Also updated README.
- Split shared logic in `public/js/common.js` into `public/js/lib/common/login.js` and `public/js/lib/config/index.js`

This PR is prepared for the upcoming refactoring. 🎉


jackycute:
Thanks @Yukaii

PR #329, opened 2017-01-23
Add fail fast option

SISheogorath:
Fail the script if a command inside it fails to prevent a install with exitcode 0 even when commands in it are failing.


SISheogorath:
Further reference: http://ss64.com/bash/set.html

Yukaii:
@SISheogorath Thanks!

PR #337, opened 2017-01-27
Upgrade-instructions: Tiny correction in README

ccoenen:
the old `npm run build` seems to be gone. `package.json` defines build:dev und build:prod, I assume they replace the old calls.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ If you are upgrading HackMD from an older version, follow these steps:
 1. Fully stop your old server first (important)
 2. `git pull` or do whatever that updates the files
 3. `npm install` to update dependencies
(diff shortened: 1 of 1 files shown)


jackycute:
@ccoenen Actually it's back again, we think old command is easier to think of.
https://github.com/hackmdio/hackmd/blob/master/package.json#L8

ccoenen:
Ah, alright :)

jackycute:
But I would like to know what do you think?

ccoenen:
I have to look it up anyway, so I have no real preference here. Both are fine.

jackycute:
OK, cool.

PR #341, opened 2017-02-03
Removed UTF-8 BOM in download function

nvsofts:
In FileSaver.js, BOM is toggleable.
(see eligrey/FileSaver.js@631047d72f4e89fab6059203ca71eb064a0bb031)


jackycute:
Oh we missed that, thanks @nvsofts !

PR #342, opened 2017-02-04
Change database config development to sqlite, test to memory

bananaappletw:
Ref: https://github.com/sequelize/express-example/blob/master/config/config.json


jackycute:
Thanks @bananaappletw

PR #343, opened 2017-02-06
Add additional comment about Prerequisite to README

sakajunquality:
`make` and `gcc` are required in [node-gyp](https://www.npmjs.com/package/node-gyp).

Without these packages, `bin/setup` ,which is instructed in Get started will fail.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,8 @@ Prerequisite
 - Node.js 4.x or up (test up to 6.7.0)
 - Database (PostgreSQL, MySQL, MariaDB, SQLite, MSSQL) use charset `utf8`
 - npm
+- gcc/make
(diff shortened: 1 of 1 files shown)


jackycute:
Thanks @sakajunquality , this might useful.

But I think it's better just reference link to prerequisite of `node-gyp` (or other dependencies).
And also we require `gcc >= 4.8` because of uWebSocket https://github.com/uWebSockets/uWebSockets#nodejs-developers
Also the phantomjs is required of PDF generating, that is `phantomjs-prebuilt` which not support all OS platforms or CPUs.

sakajunquality:
@jackycute Thank you for your comment.
I changed that part for just referencing to some packages.

jackycute:
Appreciated @sakajunquality

PR #345, opened 2017-02-07
Add minio support

SISheogorath:
Referring to https://minio.io/

I came to this as we talked about the problem with own image uploads to a self-hosted storage but keep state away from the container.

I have to **test** it later. But looks good to me.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -80,6 +80,7 @@ var imageUploadType = process.env.HMD_IMAGE_UPLOAD_TYPE || config.imageUploadTyp
 
 config.s3 = config.s3 || {}
 var s3 = {
+  endpoint: process.env.HMD_S3_ENDPOINT || config.s3.endpoint,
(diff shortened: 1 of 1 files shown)


jackycute:
Wow Minio looks fantastic!
It would be great if we support it.
Thanks @SISheogorath

mborejdo:
Adding minio support is great. I am not sure, if adding the minio-client and another case is really needed, as one could extend the S3-related configs to allow more S3-compatible storage-backends. Minio can be used with the aws-sdk (see http://docs.minio.io/docs/how-to-use-aws-sdk-for-javascript-with-minio-server)

SISheogorath:
I agree I recently found https://github.com/erasys/hackmd/commit/2a66c73b71b5a030aef04ca1fdc42ae83d14577f

I'm not sure if minio will always stay in sync with s3. But if we use AWS-sdk for now, it would smaller our dep chain which sounds very useful.

That was after building this little PR ^^

SISheogorath:
Updated the PR to use the AWS-SDK. I'll test it later.

SISheogorath:
closed in favor of #506

PR #348, opened 2017-02-09
Add default permission config

nvsofts:
Use `defaultpermission`(config.json) or `HMD_DEFAULT_PERMISSION`(environment value) to set default permission of newly created note.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ Environment variables (will overwrite other server configs)
 | HMD_USECDN | `true` or `false` | set to use CDN resources or not (default is `true`) |
 | HMD_ALLOW_ANONYMOUS | `true` or `false` | set to allow anonymous usage (default is `true`) |
 | HMD_ALLOW_FREEURL | `true` or `false` | set to allow new note by accessing not exist note url |
-| HMD_DEFAULT_PERMISSION | `freely`, `editable`, `limited`, `locked` or `private` | when creates new note, uses this permission (only applied when logged in) |
+| HMD_DEFAULT_PERMISSION | `freely`, `editable`, `limited`, `locked` or `private` | set notes default permission (only applied on signed users) |
 | HMD_DB_URL | `mysql://localhost:3306/database` | set the db url |
 | HMD_FACEBOOK_CLIENTID | no example | Facebook API client id |
 | HMD_FACEBOOK_CLIENTSECRET | no example | Facebook API client secret |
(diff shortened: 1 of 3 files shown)


jackycute:
Hi @nvsofts 
Thank you very much to send this PR.
But there already is a mechanism handling the default permission.
You might need to change that part instead of `response.js`.
https://github.com/hackmdio/hackmd/blob/master/lib/models/note.js#L516

SISheogorath:
By the way: may check https://github.com/SISheogorath/hackmd/commit/1ce93737670a02e4da7462103fd233d57e5aa400

SISheogorath:
To mention it: it's untested which is the reason why I didn't open a PR

nvsofts:
@jackycute use `model/note.js` instead of `response.js` (0a7adaf35d07efa658c040e789967acdc2eb32ff)

SISheogorath:
You still have a problem with an inconsistent database entry in case you have a typo in your default permission as you don't do a real validation. That should never happen.

jackycute:
Hmm, I think @SISheogorath was right about the validation.
@nvsofts Could you please follow this part https://github.com/SISheogorath/hackmd/commit/1ce93737670a02e4da7462103fd233d57e5aa400#diff-f7bf2b665273dfa66ffa4ad7c0a52bb9R24 to this PR?

jackycute:
Wait a sec, hold on @nvsofts 
@SISheogorath Could you explain the logic in this part? https://github.com/SISheogorath/hackmd/commit/1ce93737670a02e4da7462103fd233d57e5aa400#diff-f7bf2b665273dfa66ffa4ad7c0a52bb9R24

lib/config.js
```diff
@@ -24,6 +24,9 @@ var allowanonymous = process.env.HMD_ALLOW_ANONYMOUS ? (process.env.HMD_ALLOW_AN
 
 var allowfreeurl = process.env.HMD_ALLOW_FREEURL ? (process.env.HMD_ALLOW_FREEURL === 'true') : !!config.allowfreeurl;
 
+var defaultpermission = process.env.HMD_DEFAULT_PERMISSION || config.defaultpermission || 'editable';
```
jackycute: Please validate the permission passed to here is in the scopes.
We have 6 permission types for now.

README.md
```diff
@@ -164,6 +165,7 @@ Application settings `config.json`
 | usecdn | `true` or `false` | set to use CDN resources or not (default is `true`) |
 | allowanonymous | `true` or `false` | set to allow anonymous usage (default is `true`) |
 | allowfreeurl | `true` or `false` | set to allow new note by accessing not exist note url |
+| defaultpermission | `freely`, `editable`, `limited`, `locked` or `private` | when creates new note, uses this permission (only applied when logged in) |
```
jackycute: I would prefer following description:
`set notes default permission (only applied on signed users)`

README.md
```diff
@@ -116,6 +116,7 @@ Environment variables (will overwrite other server configs)
 | HMD_USECDN | `true` or `false` | set to use CDN resources or not (default is `true`) |
 | HMD_ALLOW_ANONYMOUS | `true` or `false` | set to allow anonymous usage (default is `true`) |
 | HMD_ALLOW_FREEURL | `true` or `false` | set to allow new note by accessing not exist note url |
+| HMD_DEFAULT_PERMISSION | `freely`, `editable`, `limited`, `locked` or `private` | when creates new note, uses this permission (only applied when logged in) |
```
jackycute: I would prefer following description:
`set notes default permission (only applied on signed users)`

nvsofts:
@SISheogorath @jackycute followed.

jackycute:
Well, that was a fast move 😄 
Nice job @nvsofts

SISheogorath:
@jackycute maybe I should stop writing everything into a single line :D

jackycute:
@SISheogorath Yeah, always consider readability even on writing code.
This might be useful: https://github.com/airbnb/javascript

PR #349, opened 2017-02-09
Fix image path problem when using filesystem backend

nvsofts:
When using `filesystem` backend in `imageUploadType` and `config.serverurl` is set, path to uploaded image is wrong.


jackycute:
Oops, thanks @nvsofts

PR #353, opened 2017-02-10
Fixed Greek mispellings and translations

stratosgear:
There were various misspellings and some translations that were not clear enough.


jackycute:
Thanks @stratosgear

PR #358, opened 2017-02-15
first envs then the config

MikeMichel:
or docker -e will not work


a60814billy:
Thanks @MikeMichel , we will merge this.

PR #359, opened 2017-02-15
Add travis ci

bananaappletw:
Add eslint syntax check by travis ci service


jackycute:
If you like to upgrade the node requirement, please also modify this line:
https://github.com/hackmdio/hackmd/blob/master/package.json#L123

bananaappletw:
@jackycute Updated.
Thanks for reminder.

a60814billy:
Thanks @bananaappletw , 
Can describe more about this PR ?

bananaappletw:
@a60814billy 
This pull request add eslint syntax check by travis ci service.
Due to [codemirror dependency](https://github.com/codemirror/CodeMirror#quickstart), upgrade Node dependency to v6.

Makefile
```diff
@@ -0,0 +1,3 @@
+lint:
+	@./node_modules/.bin/eslint .
```
a60814billy: Maybe, the script can move to package.json?

README.md
```diff
@@ -48,7 +50,7 @@ Browsers Requirement
 Prerequisite
 ---
 
-- Node.js 4.x or up (test up to 6.7.0)
+- Node.js 6.x or up (test up to 6.7.0)
```
a60814billy: In travis, already use latest node 7 version to test.
Please modify this line to node 7 latest version.

SISheogorath: I think as long as HackMD says it supports node version 6 we should test it.

.editorconfig
```diff
@@ -13,3 +13,5 @@ trim_trailing_whitespace = false
 [{.travis.yml,npm-shrinkwrap.json,package.json}]
 indent_style = space
 indent_size = 2
+[Makefile]
+indent_style = tab
```
SISheogorath: Mixing up intent styles (even for other file types) is worse and shouldn't be done!

jackycute: I'm not sure about this.
I thought different code language might have their best indent that might vary.
Would you explain your intend?

README.md
jackycute: Yeah, we'll test the 6 version.
Above comment just saying that the "test up to" should be node 7

Makefile
bananaappletw: @a60814billy For further purpose, we should use Makefile to write test.
It make clearier to write differenct test in Makefile.
Refer: https://github.com/koajs/koa/blob/master/Makefile

README.md
bananaappletw: @a60814billy I will modify this. 
Thanks.

.editorconfig
jackycute: I know the importance of consistency.
@bananaappletw what do you think?

bananaappletw: @SISheogorath @jackycute Tab character has special meaning for Makefile.
So it is proper to use tab indent instead of space in Makefile.
See [here](http://editorconfig.org/)

SISheogorath: I personally program most of the time in `vim`. And the worst thing you can have there is different intent style, than you have in your `.vimrc`.

In general speaking, you try to find a standard which says tabs or spaces. and also the intent size. Not only for you, for the whole company/project. And as long as you do not configure this in a central config file or a per file config (like possible in vim), you should stay with one type because otherwise you need to create a special config file, per project, per filetype which is more than ugly...

bananaappletw: @SISheogorath Maybe you have to take a look of [this](https://www.gnu.org/software/make/manual/make.html)

jackycute: @SISheogorath Thanks for telling more.
But I think it won't apply to Makefile, because "a tab" indent is part for the format.

SISheogorath: Yes, I was again misleading myself :D 

Thought you edit another file. It's okay :D

SISheogorath: Well, it is possible to use another character than tabs... but not really useful.

See [`.RECIPEPREFIX`](https://www.gnu.org/software/make/manual/make.html#Special-Variables)

SISheogorath:
Mhm, someone has to run `eslint --fix` before this starts to be useful, but this will possibly break all other PRs.

jackycute:
@SISheogorath yeah, so this PR is more like a pre-setup one.
We'll make the rules more fitting then fix.

Makefile
jackycute: How about combine this to webpack? since we already using that.
Seems like only koa use Makefile, maybe it's using Makefile for some other purpose than just lint or test.
I start thinking this is just a preference problem.

jackycute: And tj himself seems like stop using makefile to run npm script in his recent (an year) repos 😢.

bananaappletw: @jackycute Ok, I try to use webpack instead.
Thanks.

bananaappletw:
I change to use package.json to test lint.
Due to some third library, I add *.min.js as .eslintignore.

a60814billy:
Thanks @bananaappletw , 
we'll merge when making rules fitting our project.

jackycute:
Let's merge this first, then change the rules when needed.

Issue #129, opened 2016-05-26
不同入口的帳號合併功能 / Support merging different account data


TzuChin:
因尚未支援google登入時，
都是使用FB登入，
但現在有了google登入方式，
就開始習慣使用google登入了，
所以**希望可以把FB帳號跟google帳號合併**(應該是說多入口合併)。

雖然可以使用歷史匯入匯出的功能，
把資料搬移到google帳號底下，
但擁有者可設定的東西(例如文章權限)，
在google帳號則不能設定。

Jyny:
I have the same consider with this issue.
so rewrite the origin issue in english

When hackmd.io hasn't support sign in by google,
We are using FB to sign in.

But now hackmd.io has supported sign in by google account,
I began to get used to it.
So I hope can combined the data that signed in by FB account and google account
(or mean as multi-entry to sign in same account).

Although we can use the "import or export history" features,
to import history data to account that signed in by google account.
However, some features are lost because of it's owner actually from different account.
For an example,
the origin owner can set something (such as an article permissions),
But not In imported history's case.

Hope for the feature that support can sign in to one account by different social account.

LittleWhiteYA:
討論過後應該主要有 2 個做法
不同帳號做綁定 or 將資料全部從帳號 A 轉移到帳號 B

1. 不同帳號做綁定
優點: 不同帳號可以看到所綁定帳號的筆記和歷史紀錄，可以以其他帳號拿到"主"帳號的資料，如果綁定的帳號是不同使用者，可以達到類似 workspace 的效果
缺點: 複雜度比較高，request 的次數相比資料作全部轉移高出很多，速度會慢很多，另外以帳號綁定來達到 workspace 的效果是不良的做法，有其他 issue 會完成類似 workspace 的效果

2. 將資料全部從帳號 A 轉移到帳號 B
優點: 一次性資料轉移很單純，複雜度相對低很多，速度也快很多

jackycute:
Hi @TzuChin @Jyny and all participants
This issue might solved by a different way like above mentioned.
It's lots easier and more intuitive for users to understand.
Please comment if you have something to say.

Lin-Buo-Ren:
可以以 E-mail 地址作為識別不同的 OpenID 登入是否為同一帳號的依據嗎？
另外我附議上面 2. 的作法

bestpika:
如果用 2 的作法就變成沒辦法達到用不同帳號登入的功能…
這樣使用起來一點也不方便。

[...]

PR #370, opened 2017-02-23
不同入口的帳號合併功能 / Support merging different account data

LittleWhiteYA:
issue reference: https://github.com/hackmdio/hackmd/issues/129

I will add test later.
**For now I just finish API for transferring notes to facebook account** .
I want to check that my direction is right.

### DIFF
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -68,6 +68,12 @@ if (config.facebook) {
         clientSecret: config.facebook.clientSecret,
         callbackURL: config.serverurl + '/auth/facebook/callback'
     }, callback));
+
+    module.exports = passport.use('facebook-transferNotes', new FacebookStrategy({
+        clientID: config.facebook.clientID,
+        clientSecret: config.facebook.clientSecret,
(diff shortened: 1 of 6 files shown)


lib/auth.js
```diff
@@ -68,6 +68,12 @@ if (config.facebook) {
         clientSecret: config.facebook.clientSecret,
         callbackURL: config.serverurl + '/auth/facebook/callback'
     }, callback));
+
+    module.exports = passport.use('facebook-transferNotes', new FacebookStrategy({
```
a60814billy: Why you need `module.exports` there?
It break already exports module before.

jackycute:
@LittleWhiteYA
I would suggestion not adding that `notes-transfer-modal.ejs`. It's mostly same as `signin-modal.ejs`.
The better way is make the strings that you need to change dynamically into variables and pass them on the include, for example: `<%- include("signin-modal.ejs", { title: "example" })`.
See more here: https://github.com/mde/ejs#includes

PR #382, opened 2017-03-10
Fixed style in promo text

Himura2la:
The previous translation was really bad, I tried hard to understand why this tool will help me share knowledge ABOUT the markdown language. Then I switched to English and realized the real meaning.
Now it's much clearer


jackycute:
Thanks @Himura2la

Himura2la:
Yeas were passing...
image

PR #384, opened 2017-03-12
Webpack fixes

Yukaii:
### Changes

- limit CopyWebpackPlugin only copy files that needed
- Hide assets list in webpack build output


jackycute:
Thanks @Yukaii

PR #387, opened 2017-03-13
Extract CodeMirror instance

Yukaii:
### Changes

- Extract codemirror instance out of `public/js/index.js` to `public/js/lib/editor/index.js`
- Extract common jquery ui node out of `public/js/index` to `public/js/lib/editor/ui-elements.js`
- Extract `wrapTextWith` to `public/js/lib/editor/utils.js`
- Change `.editorconfig` to meet our new standardjs indent favor
- Patch more global variable to standardjs global variable list

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,7 +3,7 @@ root = true
 # Tab indentation
 [*]
 indent_style = space
-indent_size = 4
+indent_size = 2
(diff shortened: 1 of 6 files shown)


jackycute:
Pretty well, thanks @Yukaii

PR #417, opened 2017-04-05
Refactor editor cont.

Yukaii:
### Changes

- Seperate `line:column` and `selection` event binding
- Implement multiple event listener on the same CM event
    ```js
    const editor = new Editor()
    editor.on('cm-event', function() { console.log('first function 1') })
    editor.on('cm-event', function() { console.log('first function 2') })
    ```
   - Editor class maintain an array of functions internally, call each one on CM event sequentially
- Move `addStatusbar` method into editor class, since it's component of editor
- Drop global ui referencing in `extra.js`
- Minor coding fixes
- Create a editor config class (`public/js/lib/editor/config.js`), just a simple object store that can be shared across classes. 
   ```js
   const config = require('./lib/editor/config')
   config.abc = 'xyz'
   ```

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,11 +1,15 @@
 root = true
 
-# Tab indentation
 [*]
 indent_style = space
 indent_size = 2
 trim_trailing_whitespace = true
 insert_final_newline = true
+
+[{*.html,*.ejs}]
+indent_style = space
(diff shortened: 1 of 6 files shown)


jackycute:
Thanks @Yukaii

PR #421, opened 2017-04-11
Backend code refactoring - part 1 

a60814billy:
- [x] Extract Auth Method from app.js
- [x] Separate routing method from app.js
- [x] Refactor config.js improve readability


app.js
```diff
@@ -154,63 +141,19 @@ server.on('resumeSession', function (id, cb) {
...
-app.use(function (req, res, next) {
-  try {
-    decodeURIComponent(req.path)
-  } catch (err) {
-    logger.error(err)
-    return response.errorBadRequest(res)
-  }
-  next()
-})
+app.use(require('./lib/web/middleware/checkURiValid'))
```
Yukaii: Would it be better if we use the camel case convention `checkUrlValid` ? 😕

a60814billy: The middleware is checking uri only, I'll rename to checkURIValid

SISheogorath:
Very nice! I like the new structure!

jackycute:
Thanks @a60814billy

PR #425, opened 2017-04-17
Remove executable permission from static files

PeterDaveHello:


jackycute:
Thanks @PeterDaveHello

PR #427, opened 2017-04-17
Fix indent in views, shouldn't mix tabs with spaces

PeterDaveHello:
cc @jackycute


jackycute:
Yeah, that's true.
I think we need linter for checking `ejs` template files.

Anyway, thanks again @PeterDaveHello

PR #428, opened 2017-04-18
Move to Ubuntu Trusty on Travis CI and remove apt add-on

PeterDaveHello:
Two reasons here:

1) Ubuntu Precise is reaching EOL, though Travis CI may take care of it.
we still can move to Trusty earlier as I didn't see any dependency of
Precise here.

2) g++4.8 is built-in in Travis CI's Ubuntu Trusty environment, which
means we don't need to spend time on adding ppa repository, apt update
and apt install, could save time on the CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,9 @@
 language: node_js
+dist: trusty
 node_js:
   - 6
   - 7
   - lts/boron
 env:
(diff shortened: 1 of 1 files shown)


jackycute:
Thanks @PeterDaveHello

PR #429, opened 2017-04-18
Refactor frontend part 3

Yukaii:
### Changes

- Introduce webpack [`string-loader`](https://github.com/enmoon/string-loader), made statusbar template static. Remove related ajax code.
- Move `syncscroll.js` into `js/lib` directory. Eliminate some global variable references like `window.editor`, `window.currentMode`, and `window.syncsroll`
- Extract `appState` module, to maintain app state referencing


jackycute:
Thanks @Yukaii

PR #436, opened 2017-04-26
Fix port duplication

h8gi:
If (domain || window.location.hostname) includes port, serverurl goes wrong.


jackycute:
Hey @h8gi
Thanks for providing this PR.
Could you tell more about why `(domain || window.location.hostname)` would contain port?
And please fix you code style via `npm run test` for passing the test.

h8gi:
I'm sorry.
I wrote a port number in domain like this: `domain: 'localhost:3000'` in config.json.
This is the reason of the port duplication that happened to me.
Please ignore this pull request.

jackycute:
It's ok @h8gi 😄

Issue #418, opened 2017-04-06
Heroku deployment - deploy button causes error


VinzentE:
Hi there,

I was just trying to deploy it to heroku:
1. used the "button" (one click deploy) - but didn't work
2. just using **heroku create projectname**  => did deploy, but now not working ether.

Anyone has an exact step-by-step tutorial which I could follow? I think I am missing some point, but can not find the issue.

Would love to use the 2nd method as I think it will give me more flexibility on how to quickly change s.th. deploy it and test it.

And:
Any idea on how to make the app "serverless"? I'd love to add some add-ons, but more in a serverless environment with AWS or firebase ....

Really appreciate any ideas ... 

thanks!

--- 
log for error on heroku:
[4mSequelize [Node: 7.8.0, CLI: 2.6.0, ORM: 3.30.2][24m

Parsed url postgres://ctjxyfzpofotgd:***@ec2-23-21-96-159.compute-1.amazonaws.com:5432/det86i1phuhkhv
== 20150504155329-create-users: migrating =======
== 20150504155329-create-users: migrated (0.032s)
== 20150508114741-create-notes: migrating =======
== 20150508114741-create-notes: migrated (0.018s)
== 20150515125813-create-temp: migrating =======
== 20150515125813-create-temp: migrated (0.026s)
== 20150702001020-update-to-0_3_1: migrating =======
{ SequelizeBaseError: column "shortid" of relation "Notes" already exists
    at Query.formatError (/app/node_modules/sequelize/lib/dialects/postgres/query.js:357:14)
    at Query.<anonymous> (/app/node_modules/sequelize/lib/dialects/postgres/query.js:88:19)
    at emitOne (events.js:96:13)
    at Query.emit (events.js:191:7)
    at Query.handleError (/app/node_modules/pg/lib/query.js:143:8)
    at Connection.<anonymous> (/app/node_modules/pg/lib/client.js:180:26)
    at emitOne (events.js:96:13)
    at Connection.emit (events.js:191:7)
    at Socket.<anonymous> (/app/node_modules/pg/lib/connection.js:136:12)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:191:7)
    at readableAddChunk (_stream_readable.js:178:18)
    at Socket.Readable.push (_stream_readable.js:136:10)
    at TCP.onread (net.js:559:20)
  name: 'SequelizeDatabaseError',
  message: 'column "shortid" of relation "Notes" already exists',
  parent: 
[...]

PR #450, opened 2017-05-07
Fix heroku deployment

Yukaii:
### Changes

- Remove `postdeploy` section in heroku `app.json`.  For a clean application, there is no need to run sequelize migration.
- Remove `npm install` from heroku script. Heroku detects `yarn.lock` automatically and install package for us.


jackycute:
Thanks @Yukaii

Yukaii:
Related: #418

Issue #435, opened 2017-04-25
Image upload to S3 - ContentType not set


LluisArevalo:
Hello!

First of all, thanks for providing us hackmd.

When you are using hackmd configured with S3 to upload images, these images have no ContentType specified. By default, the image is set with the `application/octet-stream` type. When you try to open this image in a separate tab of the browser, it starts a download instead of show it.

To solve that, you have to add the correct ContentType to the parameters you are sending to S3. I have already implemented the solution in the version we forked, so let me know if you want me to create a PR to send it to you :)

Thanks again for the project!

Best,

Lluis Arévalo

jackycute:
Hey @LluisArevalo 
We'd love to see your PR.
Thanks for heads up 😄

LluisArevalo:
Thanks @jackycute, here it is: https://github.com/hackmdio/hackmd/pull/452

I hope it's well done :)

jackycute:
Well done @LluisArevalo 👍

PR #452, opened 2017-05-08
Image Upload to S3 - ContentType not set

LluisArevalo:
As discussed in the issue https://github.com/hackmdio/hackmd/issues/435, when an image is uploaded to S3, no Content-Type is set. When you try to open the image in a new tab, it starts to download the image.

With this PR, I have added a simple functionality that adds the Content-Type to the image, so the images are processed as images instead of stream objects.

It also helps to create markdowns in Hackmd and use them into Github repos showing the images correctly.

### DIFF
--- a/app.js
+++ b/app.js
@@ -548,6 +548,9 @@ app.post('/uploadimage', function (req, res) {
                 Body: buffer
               }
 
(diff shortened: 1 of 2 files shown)


jackycute:
Thanks @LluisArevalo

PR #462, opened 2017-05-17
Fix grammar and use best English standards

CrazyPython:
I'm a native speaker of both English and Chinese (My English is vastly better than my Chinese though: I can't tell you what "shirt" is in Chinese but I can converse and think in it) and I understand how different English is from Chinese...


jackycute:
Hi @CrazyPython 
Thanks for providing this PR.

However, I would like to debate for few points.
Some of wording is meant to be used as is for consistency or technical reason.
Hope you understand, I'll comment on each of them.

locales/en.json
```diff
@@ -1,15 +1,15 @@
 {
 	"Collaborative markdown notes": "Collaborative markdown notes",
-	"Realtime collaborative markdown notes on all platforms.": "Realtime collaborative markdown notes on all platforms.",
-	"Best way to write and share your knowledge in markdown.": "Best way to write and share your knowledge in markdown.",
+	"Realtime collaborative markdown notes on all platforms.": "Callaborate on markdown notes on all platforms.",
```
jackycute: So here is trying to emphasize the "realtime".
But I can see using a verb on head feels better.

locales/en.json
```diff
@@ -1,15 +1,15 @@
...
+	"Best way to write and share your knowledge in markdown.": "The best way to write and share your knowledge in markdown.",
```
jackycute: I knew there should have "The" in front of this sentence.
Because we already have the logo and "HackMD" above this.
So "The" is omitted on purpose.

locales/en.json
```diff
@@ -49,8 +49,8 @@
 	"Help": "Help",
 	"Upload Image": "Upload Image",
 	"Menu": "Menu",
-	"This page need refresh": "This page need refresh",
-	"You have an incompatible client version.": "You have an incompatible client version.",
+	"This page need refresh": "This page needs a refresh",
```
jackycute: I can't get this one.
Can the "a" be ignored?
I would prefer shorter and clear wording in most of the UI text.

locales/en.json
```diff
@@ -71,28 +71,28 @@
 	"Header": "Header",
 	"Unordered List": "Unordered List",
 	"Ordered List": "Ordered List",
-	"Todo List": "Todo List",
+	"Todo List": "Checklist",
```
jackycute: This came from GitHub Flavored Markdown (GFM) https://guides.github.com/features/mastering-markdown/
They used "Task List", so I would like to propose "Todo List".
I think this is just a preference.

locales/en.json
```diff
@@ -71,28 +71,28 @@
...
 	"Blockquote": "Blockquote",
 	"Bold font": "Bold font",
 	"Italics font": "Italics font",
 	"Strikethrough": "Strikethrough",
 	"Inserted text": "Inserted text",
-	"Marked text": "Marked text",
+	"Marked text": "Highlighted text",
```
jackycute: This came from the markdown engine we're using `markdown-it`.
You could see they use "Marked text" in their demo https://markdown-it.github.io/.
And the tag of the marked text is `<mark>` which pretty make sense.

locales/en.json
```diff
@@ -71,28 +71,28 @@
...
 	"Link": "Link",
 	"Image": "Image",
 	"Code": "Code",
 	"Externals": "Externals",
-	"This is a alert area.": "This is a alert area.",
-	"Revert": "Revert",
+	"This is a alert area.": "This is an alert area.",
+	"Revert": "Undo",
```
jackycute: This is not an undo feature.
You will revert the note to specific revision.
Undo should specifically indicate the editor "undo" action.
I would like to avoid using the same keyword on different features.

locales/en.json
```diff
@@ -1,15 +1,15 @@
...
 	"New guest note": "New guest note",
-	"Collaborate with URL": "Collaborate with URL",
-	"Support charts and MathJax": "Support charts and MathJax",
-	"Support slide mode": "Support slide mode",
+	"Collaborate with URL": "Real-time collaboration",
+	"Support charts and MathJax": "Supports charts and MathJax",
+	"Support slide mode": "Supports slide mode",
 	"Sign In": "Sign In",
-	"Below is the history from browser": "Below is the history from browser",
+	"Below is the history from browser": "Below is history from this browser",
```
jackycute: OK

locales/en.json
```diff
@@ -71,28 +71,28 @@
...
-	"This is a alert area.": "This is a alert area.",
-	"Revert": "Revert",
+	"This is a alert area.": "This is an alert area.",
+	"Revert": "Undo",
 	"Import from clipboard": "Import from clipboard",
 	"Paste your markdown or webpage here...": "Paste your markdown or webpage here...",
 	"Clear": "Clear",
 	"This note is locked": "This note is locked",
-	"Sorry, only owner can edit this note.": "Sorry, only owner can edit this note.",
+	"Sorry, only owner can edit this note.": "Sorry, only the owner can edit this note.",
```
jackycute: Can the "the" be ignored?
I would prefer shorter and clear wording in most of the UI text.

locales/en.json
```diff
@@ -71,28 +71,28 @@
...
 	"OK": "OK",
 	"Reach the limit": "Reach the limit",
-	"Sorry, you've reached the max length this note can be.": "Sorry, you've reached the max length this note can be.",
-	"Please reduce the content or divide it to more notes, thank you!": "Please reduce the content or divide it to more notes, thank you!",
+	"Sorry, you've reached the max length this note can be.": "Sorry, you've reached the maximmum length this note can be.",
+	"Please reduce the content or divide it to more notes, thank you!": "Please reduce the content or divide it into more notes, thank you!",
```
jackycute: OK

locales/en.json
```diff
@@ -71,28 +71,28 @@
...
 	"Inserted text": "Inserted text",
-	"Marked text": "Marked text",
+	"Marked text": "Highlighted text",
 	"Link": "Link",
 	"Image": "Image",
 	"Code": "Code",
 	"Externals": "Externals",
-	"This is a alert area.": "This is a alert area.",
-	"Revert": "Revert",
+	"This is a alert area.": "This is an alert area.",
```
jackycute: OK

CrazyPython:
Sure thing.

locales/en.json
CrazyPython: Hmmm... you have a point. Does adding a "in realtime" seem fine?

locales/en.json
CrazyPython: Omitting the logo generally does not mean being able to omit the "the". I've seen it many times and it always bothered me, the missing "the".

locales/en.json
CrazyPython: It can't be. The only case where you can omit a determiner ("a" or "the") is with a pronoun. "needs a" is required gere.

locales/en.json
CrazyPython: Not all of these will be todo lists. "Task List" is definitely better than "Todo List", but personally I'd prefer "checklist".

locales/en.json
CrazyPython: I've never seen the word "mark" to mean "highlighted". I think they use "marked" to refer the application of the `<mark>` tag. As the end user would not know what tag is behind, it would not make sense. And it's only `<mark>` to comply with the W3C standard that tags specify the semantics and not styling of text. (`<strike>` was superseded for this reason by `<del>`)

locales/en.json
CrazyPython: Ah. I see. Will change.

locales/en.json
CrazyPython: No, see my older explanation.

CrazyPython:
Also, good UX design says you should mark a button that deletes something "delete" instead of "Yes, do it!". I'm not sure that string is used in that context but it's useful to know.

jackycute:
I agreed with you.
Just want to be draw more attention when user doing important actions.
I would like to change the word to related action like "Yes, delete this record!".

CrazyPython:
Unrelated, but I played with styling and got this:

image

CrazyPython:
Explain `help-wanted`? Also, this is my explanation for the last edit ("Please shorten this note"): UI should tell the user what they need to do, not how to do it. In addition, "thank you" is awkward in UI.

jackycute:
Yes, help wanted.
Most of the changes are good, making the text more clear.
But some of them are related to technical terms which I can't be sure.

I would like to remove "thank you" but it should preserve "divide to several notes" to guide user for next step.

CrazyPython:
Alright.

jackycute:
I understand there're lots for thing could be improved for good UI/UX.
But actually any PR might affect whole community.
So please, give us more reasons and convince why these are better.
And I believe people would love to take.

Thanks.

locales/en.json
SISheogorath: I agree that the `the` improves the readability.

CrazyPython:
Sure.

locales/en.json
jackycute: OK

locales/en.json
jackycute: OK

locales/en.json
jackycute: Oh, so the point is at "Todo", right? Other words seem more neutral.

locales/en.json
```diff
@@ -71,28 +71,28 @@
...
 	"Clear": "Clear",
 	"This note is locked": "This note is locked",
-	"Sorry, only owner can edit this note.": "Sorry, only owner can edit this note.",
+	"Sorry, only owner can edit this note.": "Sorry, only the owner can edit this note.",
 	"OK": "OK",
 	"Reach the limit": "Reach the limit",
-	"Sorry, you've reached the max length this note can be.": "Sorry, you've reached the max length this note can be.",
-	"Please reduce the content or divide it to more notes, thank you!": "Please reduce the content or divide it to more notes, thank you!",
+	"Sorry, you've reached the max length this note can be.": "Sorry, you've reached the maximmum length this note can be.",
+	"Please reduce the content or divide it to more notes, thank you!": "Please shorten the file.",
```
jackycute: Here, please hint user they could "divide it to more notes".
Also, why do you use the word "file" than "content"?

locales/en.json
SISheogorath: Let's see what the dictionaries say: 

There is no entry for [task list](https://www.merriam-webster.com/dictionary/task%20list). So I would say its completely out.

For the word [to-do list](https://www.merriam-webster.com/dictionary/to-do%20list) there is an entry. The entry is not very detailed but at least it exists. We should only add a `-`.

The other valid option is [checklist](https://www.merriam-webster.com/dictionary/checklist) which sounds nice, is short, stays one word, and also has the most common meaning here. It also refers to the HTML elements that are used for them, checkboxes.

So the question is: Stay with `To-do list` or switch to `Checklist`

locales/en.json
CrazyPython: I'll try to change that.

I used file and and not content because this is what is typically used - *modify file*, not *modify content*.

SISheogorath:
I think we should go step by step. This PR is about language, so let's stay on topic. Further/bigger design changes should be discussed in a separated issue.

So let's fix the language quirks ;)

locales/en.json
```diff
@@ -71,28 +71,28 @@
...
 	"Paste your markdown or webpage here...": "Paste your markdown or webpage here...",
 	"Clear": "Clear",
 	"This note is locked": "This note is locked",
-	"Sorry, only owner can edit this note.": "Sorry, only owner can edit this note.",
+	"Sorry, only owner can edit this note.": "Sorry, only the owner can edit this note.",
 	"OK": "OK",
 	"Reach the limit": "Reach the limit",
-	"Sorry, you've reached the max length this note can be.": "Sorry, you've reached the max length this note can be.",
-	"Please reduce the content or divide it to more notes, thank you!": "Please reduce the content or divide it to more notes, thank you!",
+	"Sorry, you've reached the max length this note can be.": "Sorry, you've reached the maximmum length this note can be.",
```
SISheogorath: It's `maximum`.

locales/en.json
jackycute: Yeah.
But I think file is less certainly when here is only for notes.
How about just "Please shorten the note."?

SISheogorath: As we use "note" everywhere else, I agree that "note" is maybe the best word here. As an alternative to content and file, I would suggest "document".

I would like to avoid the word "file" because in the world of computers it's associated with the elements in your "explorer" and describes pictures as well as whole databases (for example in the case of SQLite).

The word "document" is used by Microsoft as well as Google and LibreOffice in their writing apps to describe the file as well as the content of it. Which makes it a good candidate, because the users already know it.

But as already mentioned to not confuse the users I would stay with "note" as this is the word everywhere else is also used.

CrazyPython: Will change!

locales/en.json
CrazyPython: Heh, I should've caught that.

locales/en.json
```diff
@@ -1,15 +1,15 @@
 {
 	"Collaborative markdown notes": "Collaborative markdown notes",
-	"Realtime collaborative markdown notes on all platforms.": "Realtime collaborative markdown notes on all platforms.",
-	"Best way to write and share your knowledge in markdown.": "Best way to write and share your knowledge in markdown.",
+	"Realtime collaborative markdown notes on all platforms.": "Collaborate on markdown notes on all platforms in realtime.",
+	"Best way to write and share your knowledge in markdown.": "<i>the simultaneous markdown editor</i>",
```
SISheogorath: Please keep the abstraction. Don't mix up translations with HTML.

And I'm not sure if that's the best change :/
I would go for `The best way to write and share your knowledge in markdown.`

CrazyPython: Can you send a PR to my PR's branch with your changes?

SISheogorath: Has landed…

CrazyPython:
@xorander00 can you do like @SISheogorath and send a PR to my PR?

locales/en.json
```diff
@@ -1,22 +1,22 @@
 {
 	"Collaborative markdown notes": "Collaborative markdown notes",
-	"Realtime collaborative markdown notes on all platforms.": "Realtime collaborative markdown notes on all platforms.",
-	"Best way to write and share your knowledge in markdown.": "Best way to write and share your knowledge in markdown.",
+	"Realtime collaborative markdown notes on all platforms.": "Collaborate on markdown notes on all platforms in realtime.",
```
SISheogorath: Should be `real-time` or `real time` -> https://www.merriam-webster.com/dictionary/real%20time

locales/en.json
SISheogorath: According to a native English speaker: "This page needs to be refreshed" is preferred.

SISheogorath:
If there are no further comments or requests, I'll test and merge this PR during this weekend. @CrazyPython can you please sign the License Change Agreement at #578?

SISheogorath:
Thanks a lot! :tada:

Issue #437, opened 2017-04-26
Support configuration of reveal.js theme


nichtich:
Looks like the reveal.js theme "black" is hardcoded in https://github.com/hackmdio/hackmd/blob/master/public/views/slide.ejs. How about a YAML configuration field to select a predefined theme:

~~~yaml
slideOptions:
  theme: white
~~~

jackycute:
Thanks @nichtich
I agree with this one.

SISheogorath:
Fixed by #471 

https://github.com/hackmdio/hackmd/blob/e5834c077f203aa5c80f777f91e0bf650614241d/public/views/slide.ejs#L32-L37

PR #471, opened 2017-05-31
add the ability to set slide theme in slide options

butlerx:
for #437

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -27,7 +27,7 @@ exports.getImageMimeType = function getImageMimeType (imagePath) {
 }
 
 exports.isRevealTheme = function isRevealTheme (theme) {
-  if (fs.existsSync(path.join(process.cwd(), 'public', 'build', 'reveal.js', 'css', 'theme', theme + '.css'))) {
+  if (fs.existsSync(path.join(__dirname, '..', 'public', 'build', 'reveal.js', 'css', 'theme', theme + '.css'))) {
     return theme
   }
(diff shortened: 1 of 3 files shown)


jackycute:
Thanks @butlerx, this looks great!

But I think we need to check if the theme is exist in the reveal.js.
https://github.com/hakimel/reveal.js/tree/master/css/theme

butlerx:
ive added a check there i wasnt really sure if doing a relative path from lib or the root dir is better though so i chose the root dir

jackycute:
@butlerx I think you did a great job.
Let's review.

lib/utils.js
```diff
@@ -23,3 +25,10 @@ exports.getImageMimeType = function getImageMimeType (imagePath) {
       return undefined
   }
 }
+
+exports.isRevealTheme = function isRevealTheme (theme) {
+  if (fs.existsSync(path.join(process.cwd(), 'public', 'build', 'reveal.js', 'css', 'theme', theme + '.css'))) {
```
jackycute: We should use `__dirname` instead of process.cwd() to prevent getting wrong path.

butlerx: `__dirname` would give `/home/user/hackmd/lib/` happy to switch  to it itll just have to be `__dirname, '..'`

jackycute: It's ok to use `..` here, lots better now.

jackycute:
Thanks @butlerx

PR #473, opened 2017-06-01
Use locally installed webpack

SISheogorath:
This makes sure webpack runs the correct version and removes the need to install webpack globally.

Based on this suggestion: https://github.com/hackmdio/docker-hackmd/pull/14#discussion_r119590376

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,8 +7,8 @@
   "scripts": {
     "test": "npm run-script standard",
     "standard": "node ./node_modules/standard/bin/cmd.js",
-    "dev": "webpack --config webpack.config.js --progress --colors --watch",
-    "build": "webpack --config webpack.production.js --progress --colors",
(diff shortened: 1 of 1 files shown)


jackycute:
I kind of agree with this, but not sure is there any possible risk of doing this.

SISheogorath:
Mhm, I did some further researches and in theory it should already work without this change: https://docs.npmjs.com/cli/run-script

> In addition to the shell's pre-existing `PATH`, `npm run` adds `node_modules/.bin` to the `PATH` provided to scripts. Any binaries provided by locally-installed dependencies can be used without the `node_modules/.bin` prefix.

jackycute:
@SISheogorath So we won't need this PR, right?

SISheogorath:
Yes, looks like we don't need it.

PR #477, opened 2017-06-03
stop the slide number being cut off

butlerx:
i noticed the slide number on reveal was being cut off this is due to the positioning assuming is position fixed, with absolute it needs to be moved in a bit

### DIFF
--- a/public/css/slide.css
+++ b/public/css/slide.css
@@ -319,6 +319,10 @@ html, body {
     position: absolute;
 }
 
+.reveal .slide-number {
+    right: 24px;
+}
+
 .print-pdf .container {
     position: relative;
     overflow-y: hidden;


jackycute:
Hi @butlerx 
Thanks for providing this PR!

Could you give few screenshots of before and after?
And is this a issue of reveal.js or it caused by our usage?

jackycute:
I think this issue might be fixed in https://github.com/hakimel/reveal.js/commit/9b11915c3a409aa456c64e9d386ac15598a4fa6b
So let's upgrade reveal.js to latest version and see.

butlerx:
its a problem with your usage of reveal since i found disabling parts of your css fixed it, but if its fixed by reveal thats grand, ive attached the images anyway, see the bottom right corner
before:
original css
after:
changed css

butlerx:
i updated reveal.js locally on master and it still seems to be a problem

jackycute:
Could you tell us your browser version and OS?

butlerx:
linux: 4.11.3-1-ARCH
Google Chrome: Version 59.0.3071.86 (Official Build) (64-bit)
Firefox: 53.0.3

jackycute:
Actually this issue was caused by one of the design in the slide mode.
We want to keep the page scrollable without showing the scrollbar.
But I can't sure on which OS or platform that the scrollbar will take place (not overlay).
So I wrote this: https://github.com/hackmdio/hackmd/blob/master/public/js/slide.js#L137
Seems the linux dist you're using also support overlay scrollbar.
Maybe I should rather to judge if it's on Windows.

butlerx:
AH right, yeah linux can be a bit of a hard target for that due to each browser being the one to decide, im going to close this as its not the solution but checking for windows seems like a better idea

jackycute:
@butlerx 
I just pushed up a commit a645f28b33405327645b1d04016c6cb74ee4a34b
Please see if that fix for you.

butlerx:
yep thats fixed it :smile:

PR #478, opened 2017-06-03
Use non-zero exit codes when build fails

SISheogorath:
Use `--bail` option to exit with a non-zero code when `webpack` fails instead ending successfully.

It also provides more useful debug output on fail.


jackycute:
Thanks @SISheogorath @Yukaii

PR #480, opened 2017-06-06
add danish translation

TheBacha:
Adding Danish translation.


jackycute:
related to #157

jackycute:
Thanks @TheBacha

PR #484, opened 2017-06-09
fix slide mode returning 500 is no slideoptions set

butlerx:
if no slideoptions were passed the rserver would return 500 if you tried to access slide mode


butlerx:
already dealt with closing pr

PR #486, opened 2017-06-14
fix s3 us-east-1 region endpoint

kota-row:
Fix Amazon S3 endpoint URL in us-east-1.

http://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region

* us-east-1(not contain `-${region}`):
```
s3.amazonaws.com
```

* other regions:
```
s3-${region}.amazonaws.com
```


jackycute:
Thanks a lot @kota-row

PR #488, opened 2017-06-16
Translate "revert" to "Zurücksetzen" in German

zuphilip:
In this way the button makes IMO much more sense:

zuruecksetzen

rather than

widerrufen

CC @janniklorenz


jackycute:
Thanks @zuphilip

PR #499, opened 2017-07-01
Use git URLs in package.json

bd808:
Using the "github:..." form to declare a dependency in package.json
makes npm attempt to install the package using an ssh clone rather than
an https clone. Some deployment environments may not allow ssh access
to external servers which will prevent the clones from succeeding. Using
the "git+https://..." form will clone the same repo from GitHub without
requiring ssh connectivity.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "start": "node app.js"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


a60814billy:
Thanks @bd808

PR #506, opened 2017-07-07
Add support for minio

marcdeop:
I know of the existence of #345 but it hasn't been updated in a while and there have been some changes on master that make that pull request useless.

I tried using our minio instance as if it were and Amazon S3 endpoint ( adding a slight change to support a different endpoint than the defaults from Amazon) but the aws-sdk refused to work ( for some reason it still contacts Amazon to verify the credentials).

Thus, I ended up opting for adding full support for minio using https://github.com/minio/minio-js

I also noticed that there is an s3bucket config option which somehow I believe it belongs to the s3 section but I didn't dare to change that.

### DIFF
--- a/lib/web/imageRouter.js
+++ b/lib/web/imageRouter.js
@@ -73,6 +73,40 @@ imageRouter.post('/uploadimage', function (req, res) {
               })
             })
             break
+
+          case 'minio':
+            var utils = require('../utils')
+            var Minio = require('minio')
+            var minioClient = new Minio.Client({
+              endPoint: config.minio.endPoint,
+              port: config.minio.port,
+              secure: config.minio.secure,
+              accessKey: config.minio.accessKey,
+              secretKey: config.minio.secretKey
+            })
+            fs.readFile(files.image.path, function (err, buffer) {
+              if (err) {
+                logger.error(err)
+                res.status(500).end('upload image error')
+                return
+              }
+
+              var key = path.join('uploads', path.basename(files.image.path))
+              var protocol = config.minio.secure ? 'https' : 'http'
+
+              minioClient.putObject(config.s3bucket, key, buffer, buffer.size, utils.getImageMimeType(files.image.path), function (err, data) {
+                if (err) {
+                  logger.error(err)
+                  res.status(500).end('upload image error')
+                  return
+                }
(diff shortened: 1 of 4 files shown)


SISheogorath:
Did you verify it works? I guess when it comes from @erasys it's running in production?

If so, can you please squash the commit messages together using `git rebase -i HEAD~3` and provide one commit message that covers everything? (This makes it easier for you and us to keep the git tree in sync)

marcdeop:
Thanks for checking this. We are indeed using it in our day to day work flow. However, I just realized that there was a minor change that I had to apply to our local setup, I will get back to you on this tomorrow evening ( European time)

lib/web/imageRouter.js
```diff
@@ -73,6 +73,40 @@ imageRouter.post('/uploadimage', function (req, res) {
...
+          case 'minio':
+            var utils = require('../utils')
+            var Minio = require('minio')
+            var minioClient = new Minio.Client({
+              endPoint: config.minio.endPoint,
+              port: config.minio.port,
+              secure: config.minio.secure,
+              accessKey: config.minio.accessKey,
+              secretKey: config.minio.secretKey
+            });
```
SISheogorath: Please remove the unneeded semicolon to fix the linter warning: https://travis-ci.org/hackmdio/hackmd/jobs/267696722#L486

marcdeop: Fixed

SISheogorath:
I was about to verify this setup and noticed that I have to think about how to set it up myself. Would it be possible to provide a little guide/documentation about how to setup minio with HackMD?

Something like what is Minio (very short), what do I need to do to set it up and what are example values for the config section.

I can probably figure it out, but not everyone is so familiar with these projects as you and me, so let's go for some docs 👍

marcdeop:
We are quite busy with some deadlines at work but I hope I will have time for this next week :-)

I will get back to you then.

SISheogorath:
Do you need help for the docs or can we make it land during the next week?

SISheogorath:
Thanks for your work! I provide the docs after merging this.

Lots of thanks for this feature! Minio will allow people to have way better self-hosted experiences.

PR #528, opened 2017-08-17
Set default permission on creation of freeurl Note

jcppkkk:
If use freeurl style to create Note, the note's permission will be `null`.
This patch try to assign default permission to freeurl Note.

### DIFF
--- a/lib/models/note.js
+++ b/lib/models/note.js
@@ -43,7 +43,10 @@ module.exports = function (sequelize, DataTypes) {
     },
     permission: {
       type: DataTypes.ENUM,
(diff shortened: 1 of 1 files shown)


jcppkkk:
I found out it's the permission column of Notes table
The Datatype should be `ENUM('freely', 'editable', 'limited', 'locked', 'protected', 'private')` (i guess) but somehow it lacks few permissions, including 'limited'.

After apply changes to permission column to Datatype `ENUM('freely', 'editable', 'limited', 'locked', 'protected', 'private')`, the note's permission works fine.

PR #529, opened 2017-08-17
Fix #420, avoid undefined content on newNote

### LINKED ISSUE #420: Error when creating notes with free url
Having `allowfreeurl` set to `true`,  an error raises:

    2017-04-11T10:51:43.884Z - error:  TypeError: Cannot read property 'length' of undefined
    at Model.<anonymous> (/private/var/projects/hackmd/hackmd/lib/models/revision.js:242:35) 

It seems that it's trying to create a revision, but as the note still doesn't have content, it fails when accessing content properties. I patched lies 240:241 with

    lastContent: note.content ? note.content : '',
    length: note.content ? note.content.length : 0,

It works for me, but i suspect it would be better that the note have a default empty content?

jcppkkk:
Fix #420 by avoid to return undefined content. 
However if other places check `undefined content` in its logic flow, this patch will have side affect.

### DIFF
--- a/lib/models/note.js
+++ b/lib/models/note.js
@@ -61,6 +61,7 @@ module.exports = function (sequelize, DataTypes) {
     },
     content: {
       type: DataTypes.TEXT,
+      defaultValue: '',
       get: function () {
         return sequelize.processData(this.getDataValue('content'), '')
       },


SISheogorath:
I personally think it's easier to simply set a default content using `sequelize`, like we do it here: 
https://github.com/hackmdio/hackmd/blob/master/lib/models/note.js#L38

lib/models/note.js
```diff
@@ -61,6 +61,7 @@ module.exports = function (sequelize, DataTypes) {
     },
     content: {
       type: DataTypes.TEXT,
+      defaultValue: '',
```
jcppkkk: `defaultValue: ''` works fine.

SISheogorath:
To keep the commit history clean, can you please run `git rebase -i HEAD~2` on this branch and squash your commit into a single one with a useful message?

Like:
```
Add default value to new note

Fixes #420 (https://github.com/hackmdio/hackmd/issues/420)
```

jcppkkk:
@SISheogorath done. :smile:

SISheogorath:
Wonderful, now it's up to @jackycute to merge this.

LGTM

SISheogorath:
@jcppkkk Can you have a look at this again?

jcppkkk:
@SISheogorath do we need a migration for default value? I mean it works fine without migrating all null content to empty content "", the default value provides empty content "" on access naturally, right?

SISheogorath:
Since it's a database change, I would say we need. I'm not completely sure how sequelize handles this internally but I would prefer that all database changes are handled by a db migrations, not by sequelize during runtime.

SISheogorath:
Fixed by #681

Issue #535, opened 2017-08-27
Remove URLs next to links when export to PDF using browser built-in print


AntonioTsai:
When I export a note to PDF by Chrome built-in print, the URL will show next to the link,
like this
![screen shot 2017-08-27 at 23 13 49](https://user-images.githubusercontent.com/10348054/29751254-75f15b00-8b7d-11e7-8673-c3f1fa5a628e.png)

The link is still available after export to PDF, so it seems we can remove the URL after the link?
And this can be done by adding CSS rule.
```css
@media print {
    a[href]::after {
        content: none !important;
    }
}
```

SISheogorath:
I would say it's a wanted behavior. 

https://github.com/hackmdio/hackmd/blob/master/public/css/extra.css#L364-L366

ccoenen:
The previous pull request was closed, since it removed all hrefs (which would have also meant: no urls in _printed_ documents). If anyone wants to go about this in a different way, feel free to propose a pull request!

SISheogorath:
For the document we want to be without printed links we can add the following at the bottom:

```
<style>
@media print {
    a[href]::after {
        content: none !important;
    }
}
</style>
```

https://hackmd.io/IYFlAYFMBMGYFoActECZ4hAY2vAnLJAEZKR7iIBmA7JGQGyVA===#

It works. I think we should provide a page (like the features page) with "Hacks" to show some of the powerful CSS features we can simply add to a document.

Nitiquita:
I can try my hand at this.

Nitiquita:
To clarify, do you want to give the user the option of having the links printed or not printed?

SISheogorath:
For me documenting the possibility of how to omit the links would be enough. But I'm open to ideas that allow toggling URL in the print-view. 

If we want to provide it via a toggle, I would go for an option in the MetaYAML section. But if there are drafts for a different solution I would welcome that, too. Buttons are more self-documenting but need more UI/UX thoughts.

ccoenen:
@Nitiquita did you get anywhere with this? If not, that's also OK, I just want to double check if anyone is working on this!

Nitiquita:
@ccoenen, sorry! I'm not working on it.

ccoenen:
No problem, no need to apologize.

AntonioTsai:
@adbahrani which version of Chrome & OS are you using?  
```html
<style> @media print { a[href]::after { content: none !important; } } </style>
```
The script works in my case.  
[...]

PR #536, opened 2017-08-28
Remove URL when export to PDF

AntonioTsai:
Mentioned in #535, I would like to remove the URLs shown next to links when export PDF using browser built-in print. The link in PDF is still available, in my personal opinion, the URL is unnecessary.

### DIFF
--- a/public/css/extra.css
+++ b/public/css/extra.css
@@ -362,6 +362,6 @@ small .dropdown a:focus, small .dropdown a:hover {
         page-break-inside: avoid !important;
     }
     a[href]:after {
-        font-size: 12px !important;
+        content: none !important;
     }
 }


SISheogorath:
I noticed two problems now:

1. PDFs created by Firefox don't put links inside
2. Of course, when you print it to paper the links will be gone. That's why printing best-practices say that you should add them this way. 

But I agree that it's not needed to add them in a PDF document and since we provide PDF export for self-hosted instances, it's exported with links anyway. I think it's something the users should decide :)

SISheogorath:
After some further research, if we really want to remove the URLs after the links I would go for this way: https://alistapart.com/article/improvingprint

Not perfect, but neither is the current way and removing the links completely breaks the paper printed versions.

AntonioTsai:
Oh, I only tested it on Chrome and Safari, and I forgot the situation when printing it to paper.

ccoenen:
Thanks for the pull request, but I think this will require a different approach.

PR #538, opened 2017-08-29
Update yarn.lock

madebyherzblut:
The current `yarn.lock` is out of sync with the `package.json` which in turn results in failed deployments to Heroku.


jackycute:
Thanks @madebyherzblut

edoardoo:
I think this problem has surfaced again. I got the same inconsistency deploying on heroku.

PR #541, opened 2017-09-04
Fix naming typo.

Stonesjtu:
Newbie to this repo, but I don't think `facebook` is a good name prefix here. If any naming dependency is missing, reject it.


jackycute:
Thanks @Stonesjtu @SISheogorath

PR #550, opened 2017-09-22
Fix broken profile images in GitLab

### LINKED ISSUE #549: GitLab OAuth only works with gravatar enabled
Hi, 

we just noticed that OAuth via GitLab only works correctly if Gravatar support is enabled in GitLab.

*Requirements:*  
* HackMD 0.5.1
* GitLab CE 9.5.5 (with Gravatar support disabled)

*Expected behaviour:*  
Login via GitLab username and password via "Sign In" button. Username is shown in the upper right of HackMD (replacing the login button).

*Actual behaviour:*  
Login via GitLab seems to succeed. Upper right button still shows "Sign In".

*Steps to reproduce:*  
1. Configure OAuth via GitLan as described in the documentation
1. Disable Gravatar support in GitLab
1. Login to HackMD via GitLab
    The login seems to succeed but you will still see the login button instead of the username
1. HackMD shows the following error
    `error: read me failed: TypeError: Cannot read property 'replace' of null`

After (re)enabling Gravatar support in GitLab login via OAuth succeeds without any errors.

SISheogorath:
This PR should fix #549 

Since I don't run a private GitLab Instance to test this, it would be nice, to be tested by the community.

This PR is compatible with `0.5.1`.

# How to build?

```
git clone https://github.com/SISheogorath/hackmd-docker
cd hackmd-docker
$EDITOR docker-compose.yml

docker-compose up
```

In the compose file, please uncomment the [build section](https://github.com/SISheogorath/hackmd-docker/blob/master/docker-compose.yml#L21-L23) and change it to the following:

```
    build: 
      context: .
      dockerfile: debian/Dockerfile
      args:
         - "VERSION=fix/gitlabAvatar"
         - "HACKMD_REPOSITORY=https://github.com/SISheogorath/hackmd.git"
```

And add the needed environment variables for the gitlab config to the [environment section](https://github.com/SISheogorath/hackmd-docker/blob/master/docker-compose.yml#L25-L31).

This should allow you to validate the changes.

### DIFF
--- a/lib/models/user.js
+++ b/lib/models/user.js
@@ -104,8 +104,12 @@ module.exports = function (sequelize, DataTypes) {
             break
           case 'gitlab':
             photo = profile.avatarUrl
-            if (bigger) photo = photo.replace(/(\?s=)\d*$/i, '$1400')
-            else photo = photo.replace(/(\?s=)\d*$/i, '$196')
+            if (photo) {
+              if (bigger) photo = photo.replace(/(\?s=)\d*$/i, '$1400')
+              else photo = photo.replace(/(\?s=)\d*$/i, '$196')
+            } else {
+              photo = letterAvatars(profile.username)
+            }
             break
           case 'dropbox':
             // no image api provided, use gravatar


smarrk:
I patched this PR manually against `0.5.1` and tested it successfully with enabled/disabled Gravatar support in GitLab (`9.9.5`) and set/unset profile picture.

Profile picture set:
HackMD always shows the picture set by GitLab (regardless the status of Gravatar support).

Profile picture not set:
HackMD shows a Letter-Avatar when Gravatar support is disabled and the autogenerate picture (by GitLab) when Gravatar support is enabled.

SISheogorath:
Nice! Thanks a lot!

Now we only have to wait for someone with write access

PR #562, opened 2017-10-07
Fix LDAP problem about missing uidNumber

### LINKED ISSUE #432: If uidNumber does not exist, session management fails with LDAP authentication.
Hello, I am KONISHI Keisuke.

Thank you for providing HACKMD.

I built HackMD with Docker, but LDAP authentication did not work properly. (No problem if there is one user)

Within Hackmd, in the case of LDAP authentication, the uidNumber attribute of LDAP is fixed as profile.id. However, if uidNumber does not exist in LDAP, profile.id will always be "LDAP-undefined".
Because the profile.id is the same, all the sessions connected by LDAP are changed to the user who last logged in with LDAP.
(Private documents will be seen!)

In LDAP, uidNumber is not mandatory.
For this reason, we need to fix HackMD.
As a workaround, there is a method to make it possible to set arbitrary attributes in environment (for example, HMD_LDAP_PROFILE_ID_ATTRIBUTE), or to use the value of existing HMD_LDAP_SEARCHFILTER (after setting username) for profile.id.

auth.js
`      var profile = {
        id: 'LDAP-' + user.uidNumber,
        username: user.uid,
        displayName: user.displayName,
        emails: user.mail ? [user.mail] : [],
        avatarUrl: null,
        profileUrl: null,
        provider: 'ldap'
      }
`

SISheogorath:
Try to fix #432

After checking multiple RFCs and references `uid` and `sAMAccountName` seem to be the most commonly used fields.

`uidNumber` is a POSIX reference for the uid that is used when people log in to a Unix host. This should not be used for HackMD.

I don't think that it's a perfect solution. My first try was to add a settings field for it, but it seems to overcomplicate things and is unnecessary long.

I'll test this later. Hopefully during this weekend.

### DIFF
--- a/lib/web/auth/ldap/index.js
+++ b/lib/web/auth/ldap/index.js
@@ -23,9 +23,10 @@ passport.use(new LDAPStrategy({
     tlsOptions: config.ldap.tlsOptions || null
   }
 }, function (user, done) {
+  var uuid = user.uidNumber || user.uid || user.sAMAccountName
   var profile = {
-    id: 'LDAP-' + user.uidNumber,
-    username: user.uid,
+    id: 'LDAP-' + uuid,
+    username: uuid,
     displayName: user.displayName,
     emails: user.mail ? [user.mail] : [],
     avatarUrl: null,


ccoenen:
I have no way of testing this change, I'm afraid. I'm paging @alecdwm who may be more familiar with this part of the code.

SISheogorath:
I tested this code. It works and fixes the problem.

The remaining question is if a more generic solution is wanted/needed.

@komikoni  @mborejdo @traedamatic

SISheogorath:
Added `user.uidNumber` to not break existing setups.

Issue #493, opened 2017-06-20
[Design suggestion] Make "New note" pop


tjohnell-handy:
Yes, pop. I know it's a trigger word, but in this case I think it really is required.  I hunted on the main page for several seconds trying to figure out how the heck I create a new note:

![image](https://user-images.githubusercontent.com/13623283/27354879-22b31216-55d7-11e7-8682-c36100c37128.png)

Seems like it should be a large green/button, front and center.

geekyd:
I would like to work on this. Please assign this issue to me.

ccoenen:
It appears I can only assign an issue to people that are formally "on the team". Please feel free to work on this issue :-)

PR #564, opened 2017-10-07
Adds green color to new note button

geekyd:
Changes: 
- Adds green color to new note button for better visibility 

Screenshot:
screenshot from 2017-10-07 19-25-42

Fixes: #493

### DIFF
--- a/public/views/index/body.ejs
+++ b/public/views/index/body.ejs
@@ -13,14 +13,14 @@
                             </li>
                             <div class="ui-signin" style="float: right; margin-top: 8px;<% if(signin) { %> display: none;<% } %>">
                                 <% if(allowAnonymous) { %>
-                                <a type="button" href="<%- url %>/new" class="btn btn-sm btn-link"><i class="fa fa-plus"></i> <%= __('New guest note') %></a>
+                                <a type="button" href="<%- url %>/new" class="btn btn-sm btn-primary"><i class="fa fa-plus"></i> <%= __('New guest note') %></a>
                                 <% } %>
                                 <% if(facebook || twitter || github || gitlab || dropbox || google || ldap || email) { %>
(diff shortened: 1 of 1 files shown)


public/views/index/body.ejs
```diff
@@ -13,14 +13,14 @@
                             </li>
                             <div class="ui-signin" style="float: right; margin-top: 8px;<% if(signin) { %> display: none;<% } %>">
                                 <% if(allowAnonymous) { %>
-                                <a type="button" href="<%- url %>/new" class="btn btn-sm btn-link"><i class="fa fa-plus"></i> <%= __('New guest note') %></a>
+                                <a type="button" href="<%- url %>/new" class="btn btn-sm btn-success"><i class="fa fa-plus"></i> <%= __('New guest note') %></a>
```
SISheogorath: I'm not sure if we should use this for not-singed in people, too. As you can see below the sign-in button already uses `btn-success` and from my perspective, there shouldn't be two buttons with this color

SISheogorath:
In generell how about using `btn-primary` instead of `btn-success` for creating new notes? I'm not sure about the UX for the front page when you are a not signed in person. Because the "all features" button uses this color, too. But in general, it's maybe useful because it bites with the sign-in button.

public/views/index/body.ejs
geekyd: @SISheogorath The purpose of making these changes is to highlight the `new note` button. Now the button is green, which makes it easier for the user to spot it.

geekyd:
@SISheogorath Using `btn-primary` seems a better choice.

public/views/index/body.ejs
geekyd: @SISheogorath I have changed the color please review.

screenshot from 2017-10-07 23-21-34

ccoenen:
Thanks for taking the time!

SISheogorath:
Updated Screenshot:

image


I like it! 👍

PR #566, opened 2017-10-09
createdAt DESC with quotation marks did not work with MySQL fixes #565

### LINKED ISSUE #565: History (Version list) unsorted?
When I try to go over recent changes, the list looks like this for me: 

grafik

It appears not to be sorted by anything meaningful. It's not sorted by date, not by size of the document (not that this would make sense).

The Version in use is 0.5.1, running on MySQL.

ccoenen:
This _has_ to be doublechecked with PostgreSQL! I did not yet test that.

### DIFF
--- a/lib/models/revision.js
+++ b/lib/models/revision.js
@@ -110,7 +110,7 @@ module.exports = function (sequelize, DataTypes) {
           where: {
             noteId: note.id
           },
-          order: '"createdAt" DESC'
+          order: [['createdAt', 'DESC']]
         }).then(function (revisions) {
           var data = []
           for (var i = 0, l = revisions.length; i < l; i++) {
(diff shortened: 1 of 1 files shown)


SISheogorath:
Sadly as it turned out, the quotes are needed for PostgreSQL to work. Otherwise, the handling of the column name is case-insensitive and PostgreSQL can't find the column anymore, which breaks the functionality completely.

But using the array syntax (`order: [['createdAt', 'DESC']]`) seems to work.

ccoenen:
How about Backticks? They are used all over the MySQL documentation?

ccoenen:
(Sequelize docs on ordering: http://docs.sequelizejs.com/manual/tutorial/querying.html#ordering )

SISheogorath:
That's for version 4.

We use version 3. See https://sequelize.readthedocs.io/en/v3

ccoenen:
Thanks! http://sequelize.readthedocs.io/en/v3/docs/querying/#ordering

ccoenen:
So. if anyone could take this updated commit for a spin on PostgreSQL, we could probably merge it into master :)

PR #567, opened 2017-10-09
converting all content fields to MEDIUMTEXT (affects MySQL only)

### LINKED ISSUE #521: silent data loss for pads above 64KB on MySQL/MariaDB
# Problem

Pads above 64KB get truncated silently (on MySQL / Mariadb)
The same is true for their revisions.

# Desired Solution (Preferred)

The Field should deal with larger Texts. 64KB is not extremely much (though it _is_ quite long). `MEDIUMTEXT` (16MB) or `LONGTEXT` (4GB) would give the user more headroom.

# Desired Solution (Additionally)

The editor should warn in case of data loss, regardless of where the limit lies.

# Problem source

[content](https://github.com/hackmdio/hackmd/blob/d1d6d5810b12645ddb02275ce0c2498b2189a8a0/lib/models/note.js#L63) is [`Sequelize.TEXT`](http://docs.sequelizejs.com/variable/index.html#static-variable-DataTypes).

This becomes
```
MariaDB [hackmd]> show fields from Notes;
+------------------+----------------------------------------------+------+-----+---------+-------+
| Field            | Type                                         | Null | Key | Default | Extra |
+------------------+----------------------------------------------+------+-----+---------+-------+
| id               | char(36)                                     | NO   | PRI |         |       |
| shortid          | varchar(255)                                 | NO   | UNI | NULL    |       |
| alias            | varchar(255)                                 | YES  | UNI | NULL    |       |
| permission       | enum('freely','editable','locked','private') | YES  |     | NULL    |       |
| viewcount        | int(11)                                      | NO   |     | 0       |       |
| title            | text                                         | YES  |     | NULL    |       |
| content          | text                                         | YES  |     | NULL    |       | <----
| authorship       | text                                         | YES  |     | NULL    |       |
| lastchangeAt     | datetime                                     | YES  |     | NULL    |       |
| savedAt          | datetime                                     | YES  |     | NULL    |       |
| createdAt        | datetime                                     | NO   |     | NULL    |       |
| updatedAt        | datetime                                     | NO   |     | NULL    |       |
| deletedAt        | datetime                                     | YES  |     | NULL    |       |
| ownerId          | char(36)                                     | YES  |     | NULL    |       |
| lastchangeuserId | char(36)                                     | YES  |     | NULL    |       |
+------------------+----------------------------------------------+------+-----+---------+-------+
15 rows in set (0.00 sec)
```

MySQL `text` (no surprise here).

[TEXT types](https://dev.mysql.com/doc/refman/5.7/en/blob.html) in combination with [Storage Requirements](https://dev.mysql.com/doc/refman/5.7/en/storage-requirements.html#data-types-storage-reqs-strings) leads me to believe that `TEXT` is `2^16` bytes large - or ~64KB.

ccoenen:
This fixes #521 by making the `TEXT` column larger for MySQL / MariaDB. In PostgreSQL, `TEXT` is already very large, but on MySQL this is just ~64KiB long.

Includes a migration to enable this. This should not change anything on PostgreSQL, as Sequelize maps the field to `TEXT` on that platorm anyway.

Can somebody please confirm that this does not break (or in fact change) PostgreSQL? You'll need to run 

`node_modules/.bin/sequelize db:migrate` (perhaps with your environment)

### DIFF
--- a/lib/migrations/20171009121200-longtext-for-mysql.js
+++ b/lib/migrations/20171009121200-longtext-for-mysql.js
@@ -1,0 +1,16 @@
+'use strict'
+module.exports = {
+  up: function (queryInterface, Sequelize) {
+    queryInterface.changeColumn('Notes', 'content', {type: Sequelize.TEXT('long')})
+    queryInterface.changeColumn('Revisions', 'patch', {type: Sequelize.TEXT('long')})
+    queryInterface.changeColumn('Revisions', 'content', {type: Sequelize.TEXT('long')})
+    queryInterface.changeColumn('Revisions', 'latContent', {type: Sequelize.TEXT('long')})
+  },
+
+  down: function (queryInterface, Sequelize) {
+    queryInterface.changeColumn('Notes', 'content', {type: Sequelize.TEXT})
+    queryInterface.changeColumn('Revisions', 'patch', {type: Sequelize.TEXT})
+    queryInterface.changeColumn('Revisions', 'content', {type: Sequelize.TEXT})
+    queryInterface.changeColumn('Revisions', 'latContent', {type: Sequelize.TEXT})
+  }
+}
--- a/lib/models/note.js
+++ b/lib/models/note.js
@@ -60,7 +60,7 @@ module.exports = function (sequelize, DataTypes) {
       }
     },
     content: {
-      type: DataTypes.TEXT,
+      type: DataTypes.TEXT('long'),
       get: function () {
         return sequelize.processData(this.getDataValue('content'), '')
       },
--- a/lib/models/revision.js
+++ b/lib/models/revision.js
@@ -58,7 +58,7 @@ module.exports = function (sequelize, DataTypes) {
       defaultValue: Sequelize.UUIDV4
     },
     patch: {
-      type: DataTypes.TEXT,
+      type: DataTypes.TEXT('long'),
       get: function () {
         return sequelize.processData(this.getDataValue('patch'), '')
       },
@@ -67,7 +67,7 @@ module.exports = function (sequelize, DataTypes) {
       }
     },
     lastContent: {
-      type: DataTypes.TEXT,
+      type: DataTypes.TEXT('long'),
       get: function () {
         return sequelize.processData(this.getDataValue('lastContent'), '')
       },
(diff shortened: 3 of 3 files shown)


SISheogorath:
I'll provide some instructions for someone who want to validate this version with PostgreSQL.

For MySQL a few additonal changes are needed.

# How to build?

```
git clone https://github.com/SISheogorath/hackmd-docker
cd hackmd-docker
$EDITOR docker-compose.yml

docker-compose up
```

In the compose file, please uncomment the [build section](https://github.com/SISheogorath/hackmd-docker/blob/master/docker-compose.yml#L21-L23) and change it to the following:

```
    build: 
      context: .
      dockerfile: debian/Dockerfile
      args:
         - "VERSION=ix-mysql-text-length"
         - "HACKMD_REPOSITORY=https://github.com/ccoenen/hackmd.git"
```

And add the needed environment variables for the gitlab config to the [environment section](https://github.com/SISheogorath/hackmd-docker/blob/master/docker-compose.yml#L25-L31).

This should allow you to validate the changes.

ccoenen:
travis fail was due to JS Style (Semicolons which it did not like. I pushed a new commit.

SISheogorath:
On PostgreSQL the migration shows a little warning:

```
>> WARNING: PostgreSQL does not support TEXT with options. Plain `TEXT` will be used instead. 
>> Check: http://www.postgresql.org/docs/9.4/static/datatype.html
```

But that can be ignored.

ccoenen:
Yes, that's correct. `TEXT` in PostgreSQL is much larger (1GB) than `MEDIUMTEXT` in MySQL (16MB).

Unless somebody has a problem with "a warning is being generated", I'd merge this. (Yes, I'm aware that a warning is not very nice, but it's certainly nicer than data loss).

SISheogorath:
Let's give hand it to  @jackycute. Maybe he has an opinion on it.

jackycute:
I think it's ok to change it to medium text type, but it will be better if the size of this type is not limited, it should grow with the data, or it will meet the veil anyways.

ccoenen:
I changed all fields to be `LONGTEXT`

SISheogorath:
LGTM.

But I would like to keep it out of master until we plan to release. Reason: We currently tell people to use master instead of 0.5.1 when they install, migrate or upgrade their setup and they maybe don't run a `node_modules/.bin/sequelize db:migrate` when they do a `git pull`.

@ccoenen are you fine with this?

ccoenen:
Perfectly fine for me.

PR #569, opened 2017-10-09
Provide table for permissions

SISheogorath:
Should help to visualize the permission settings.

### DIFF
--- a/public/docs/features.md
+++ b/public/docs/features.md
@@ -47,12 +47,15 @@ or import content from your **clipboard** <i class="fa fa-clipboard"></i>, and t
 It is possible to change the access permission to a note through the little button on the top right of the view.
 There are four possible options:
 
-<i class="fa fa-leaf fa-fw"></i> **Freely**: Anyone can edit this note.
-<i class="fa fa-pencil fa-fw"></i> **Editable**: A signed-in user can edit this note.
-<i class="fa fa-id-card fa-fw"></i> **Limited**: People have to sign-in to view and edit this note.
-<i class="fa fa-lock fa-fw"></i> **Locked**: Anyone can view this note but only the owner can edit it.
-<i class="fa fa-umbrella fa-fw"></i> **Protected**: People have to sign-in to view this note but only owner can edit.
-<i class="fa fa-hand-stop-o fa-fw"></i> **Private**: Only the owner can view and edit this note.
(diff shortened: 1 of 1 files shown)


public/docs/features.md
```diff
@@ -47,12 +47,15 @@ or import content from your **clipboard** <i class="fa fa-clipboard"></i>, and t
...
-<i class="fa fa-lock fa-fw"></i> **Locked**: Anyone can view this note but only the owner can edit it.
-<i class="fa fa-umbrella fa-fw"></i> **Protected**: People have to sign-in to view this note but only owner can edit.
-<i class="fa fa-hand-stop-o fa-fw"></i> **Private**: Only the owner can view and edit this note.
+|                        |Owner read/write|Signed-in read|Signed-in write|Guest read|Guest write|
+|-----------------------:|:--------------:|:------------:|:-------------:|:--------:|:---------:|
+|**Freely** <i class="fa fa-leaf fa-fw"></i>                                          |✔|✔|✔|✔|✔|
+|**Editable** <i class="fa fa-pencil fa-fw"></i>                                      |✔|✔|✔|✔|✖|
+|**Limited** <i class="fa fa-id-card fa-fw"></i>                                      |✔|✔|✔|✖|✖|
+|**Locked** <i class="fa fa-lock fa-fw"></i>                                          |✔|✔|✖|✔|✖|
+|<a style="white-space:nowrap;">**Protected** <i class="fa fa-umbrella fa-fw"></i></a>|✔|✔|✖|✖|✖|
```
ccoenen: this one has an extra a-tag. Is this intentional? If so, why not a semantically neutral `<span>`?

SISheogorath: Main reason was that `<a></a>` vs `<span></span>` has 6 char less :D 

I should change it. it looked a bit ugly in HackMD but in the normal editor, it should be fine.

The reason for the tag itself is that otherwise, the line would break (The umbrella would jump into the next line). In the [original document](https://hackmd.io/EwNg7CIBwGYwtFARgQwCbwCwE4bfktgMYICmRADAMywCMYKFYYQA) I added the document style underneath. But I want to avoid this for the Features document.

SISheogorath:
Changed that and swapped the icons into the front. Reason: The permission settings look exactly like that also I think it's more mobile friendly because you end up with vertical scrolling and the space in front looks wasted.

SISheogorath:
Updated to use the bootstrap class instead of inline styling.

PR #570, opened 2017-10-10
Link constants.js to ./tmp for heroku deployments

### LINKED ISSUE #500: Deploying to Heroku Problems :(
Hi guys, 

We have been having problems running hackmd in production since a few weeks ago. We have a deployed instance in heroku that runs with cedar-16 stack (node v8+) and there are two main issues that kept us from updating hackmd:

The first one is [an error in UWS at boottime](https://github.com/uNetworking/bindings/issues/37) which has to do with UWS not having the right compiled version for node's latest ABI (57).

The second one has to do with [this line](https://github.com/hackmdio/hackmd/blob/master/app.js#L40):

```
ejs.renderFile(constpath, data, {}, function (err, str) {
  if (err) throw new Error(err)
  fs.writeFileSync(path.join(__dirname, './public/build/constant.js'), str)
})
```

Seems like when you start hackmd, it will create a file named `./public/build/constant.js` that contains something along the lines of:

```
window.domain = 'your.domain.com'
window.urlpath = ''
window.debug = false
window.version = '0.5.1'

window.GOOGLE_API_KEY = ''
window.GOOGLE_CLIENT_ID = ''
window.DROPBOX_APP_KEY = ''
```

The problem is that, as you may know, heroku's file system is READ-ONLY. Therefore I can't start my server because when it tries to start it in heroku, `writeFileSync` will throw an exception.

I'd suggest wrapping `fs.writeFileSync(path.join(__dirname, './public/build/constant.js'), str)` in a try...catch block and managing this case. Seems to me that this way of passing data to all of the views needs to be refactored, but i'm not familiar with the codebase enough to suggest how.

Thankks guys!!

SISheogorath:
Fixes #500 

It would be nice when someone can verify that this fixes the Heroku deployments.

### DIFF
--- a/bin/heroku
+++ b/bin/heroku
@@ -28,4 +28,8 @@ EOF
 
   # build app
   npm run build
+
+  # Link "constant.js" to the writable temp directory for Heroku deployments
+  ln -s ./tmp/constant.js ./public/build/constant.js
 fi
+
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "dev": "webpack --config webpack.config.js --progress --colors --watch",
     "build": "webpack --config webpack.production.js --progress --colors --bail",
     "postinstall": "bin/heroku",
-    "start": "node app.js"
+    "start": "ls -la ./public/build/ && node app.js"
   },
   "dependencies": {
     "Idle.Js": "git+https://github.com/shawnmclean/Idle.js",


jackycute:
@f3r Could you help to test this PR?

SISheogorath:
This can not fix the problem.

PR #574, opened 2017-10-10
Add "Table of Contents" in README.md

PeterDaveHello:


SISheogorath:
Can we restructure the README a bit with this PR?

Like adding some additional Headlines like Providers (Heroku, Docker, …), Setup (Requirements, Getting Started, Upgrade, Migrations, …) and Configuration.

I think it would also be nice to have the "Getting Started" part at the top instead of the middle of the README.

PeterDaveHello:
To confirm, so are there 4 headlines in total you'd like to have?

SISheogorath:
I was more thinking about something like this:

- Getting Started
  - native setup
  - Heroku
  - Docker
- Upgrade
  - App
  - Migration to 0.5.0
  - Migration to 0.4.0
…

and so on. So we have not simply a list of topics. Instead, it has some logical structure.

PeterDaveHello:
hmmm, if it's not a too trivial solution and not directly related, maybe separate it in a later PR?

SISheogorath:
Mhm. I see the purpose of this PR but I don't think it's as helpful as it could be. As a shortcut to the sections the TOC is useful, for sure, but I don't think that it's really useful for newcomers in the current README structure.

When I check out a project I try to find the getting started section (would expect it on top of everything else). When I find a list of links as TOC that less attractive than some kind of place where I get started instantly, which is right now the docker section. Means we should either rename it or restructure the document when we are about to visualize the structure.

PeterDaveHello:
That's true, my concern is that I'm not sure what would you like it to be, wanna share more details?

SISheogorath:
Mhm I would recommend a structure like this:

- Installation
  - Getting started (Native install)
    - Prerequisite
    - Instructions
  - Heroku Deployment
  - HackMD by docker container
- Upgrade
  - Native setup
  - Heroku Deployment
  - HackMD by docker container
- Configuration
 - Environment variables (will overwrite other server configs)
 - Application settings `config.json`
 - Third-party integration api key settings
 - Third-party integration oauth callback urls
- Developer Notes
  - Structure
  - Operational Transformation

---

But I'm open to other ideas. The upgrade section for docker and Heruko do currently not exist and need to be written. Also, the Headlines of some sections are renamed, like "docker-hackmd".

The Upgrade section for Docker should be something like this:

> For detailed updated instruction please check the [instructions in the docker-hackmd repository](https://github.com/hackmdio/docker-hackmd#update)

For Heruko probably the link to the Deployment section is enough:

> See [Heroku deployment section](#heroku-deployment)

I think that should be possible. Feel free to surprise with more ideas :)

PeterDaveHello:
@jackycute is this good with you?

jackycute:
@PeterDaveHello Yes, it looks awesomesoup.

jackycute:
@PeterDaveHello  Oh, but I think we can keep the "Prerequisite" as is.

PeterDaveHello:
Okay I'll reorganize as @SISheogorath requested and then add the Table of Contents.

PeterDaveHello:
Not sure if I'm right or wrong, @SISheogorath I adjust few points as I thought some of them may not really match the content in current README.md, and I'm trying to not touch too much content in this PR as the content is not the main point here, would you please take a look if current version good with you? Thanks.

PeterDaveHello:
To prevent to overhead of manual updating the Table of Contents, I integrate the [`doctoc`](https://www.npmjs.com/package/doctoc) package to help update it, would be useful in future updates!

SISheogorath:
Is `doctoc` already in packages.json as dev dependency?

PeterDaveHello:
@SISheogorath my fault, thanks for that catching! Updated now!

PeterDaveHello:
Let me know if there is anything I should do please :)

PeterDaveHello:
Yeah I feel the same thing, just try not to mix them :)

package.json
```diff
@@ -151,6 +152,7 @@
     "babel-runtime": "^6.20.0",
     "copy-webpack-plugin": "^4.0.1",
     "css-loader": "^0.26.1",
+    "doctoc": "^1.3.0",
```
ccoenen: doctoc adds a lot of dependencies, do we really want this?
http://npm.anvaka.com/#/view/2d/doctoc

SISheogorath: If it's used to generate the ToC it should be included in the development dependencies like `webpack` and `standard` and everything else we use to generate code from or for the repository.

https://github.com/hackmdio/hackmd/blob/5ce8f40eac7fccfcfca175cf79ccfa86a84bc113/package.json#L144-L172

Are you worried about the production machines who run HackMD and need to install` doctoc` because `webpack` and other dependencies are needed for the build process?

ccoenen: webpack is used by everyone who runs their own server
standard is run by evyone changing any line of javascript
doctoc would only be needed for those updating the readme. I wonder if this would not be a candidate for `optionalDependencies`?

SISheogorath: I wouldn't say that it's an optionalDependency since they are installed by `npm install` if you don't specify `--no-optional`.

The devDependency description says:

> If someone is planning on downloading and using your module in their program, then they probably don't want or need to download and build the external test or documentation framework that you use.

`doctoc` sounds like a documentation tool to me.

No matter if it is or not, I think when we want to keep the comments about doctoc, we should add it to the dependency list. If not, we should remove the comments.

@PeterDaveHello is your `doctoc` installed globally or did you install it in the project scope?

PeterDaveHello: I use it in the project scope, not everyone knows where it comes from, so the dependency can help anyone to fetch it without the issue.

PeterDaveHello:
Anything else I can do here?

SISheogorath:
If no further comments, I'll merge this tomorrow.

Adding `doctoc` to package.json as dev dependency seems fine to me since it's for documentation reasons.

PR #586, opened 2017-10-13
Add jsonlint script to ensure all json files are valid

PeterDaveHello:
Wanna add one more test  :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,8 @@
   "main": "app.js",
   "license": "MIT",
   "scripts": {
-    "test": "npm run-script standard",
+    "test": "npm run-script standard && npm run-script jsonlint",
+    "jsonlint": "find . -not -path './node_modules/*' -type f -name '*.json' | while read json; do echo $json ; jsonlint -q $json; done",
     "standard": "node ./node_modules/standard/bin/cmd.js",
     "dev": "webpack --config webpack.config.js --progress --colors --watch",
     "build": "webpack --config webpack.production.js --progress --colors --bail",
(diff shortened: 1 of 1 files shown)


SISheogorath:
How about having a look at code climate?

it should allow all these kinds of testing :) I played around with it here:
https://github.com/sisheogorath/hackmd/tree/qa/codeclimate

You are very welcome to extend and make it usable 😄

PeterDaveHello:
Looks cool, but I may need some time to learn about it, looks like the branch needs more tuning to be ready? If I'm not wrong, maybe migrate to code climate then? Let me know what do you think, thanks!

SISheogorath:
Mhm maybe it's not the worst idea, to add this PR right now and introducing codeclimate later since it will need a lot of work and as it'll touch almost all files, I would like to do it after the AGPL change.

---

For this PR 👍  We should add this. Can you add it to the test command as well, so we validate both using travis?

PeterDaveHello:
@SISheogorath sorry that i didn't propose first, I have another branch to refactor the Travis CI build:
https://github.com/PeterDaveHello/hackmd/tree/refactor-TravisCI-config

which will introduce matrix syntax and than add multiple tests in(not mean to earn many commits, I add them all in a single one), since current one needs few tuning, so that I didn't do it all together, would that be great for you?

SISheogorath:
i was referring to this test command:

https://github.com/hackmdio/hackmd/blob/53c2d0b5ca5901c1d1cad819e2049b16fba18ea8/package.json#L7

:+1: for the Travis refactoring

PeterDaveHello:
You mean to combine it like  `"test": "npm run-script standard && npm run-script jsonlint", `? I wonder if it'll be too long and mixed with too many things especially if we're going to add more tests?

SISheogorath:
Yes, that's what I meant.

We will probably move it to a more useful place later but for now, I think that's the place to go.

PeterDaveHello:
Done.

PR #578, opened 2017-10-11
License Change Agreement

jackycute:
We [propose a license change, from **MIT** to **AGPL 3.0**](https://hackmd.io/s/r1_4j9_hZ), for the open source project HackMD Community Edition (CE), and ask for your understanding and agreement.
Please leave a comment below the PR saying: 

**“I agree to the change of license from MIT to AGPL on all my past code contribution to the project HackMD. 
Signed-off-by:  Full_name <my_real@mail.address>“**

Thank you again for your contribution to the project and the community.
Please let us know if you have any questions.

PR #589, opened 2017-10-15
Adds enable/disable PDF export via config

### LINKED ISSUE #561: Disable PDF export by config
The PDF export is currently experimental and causes problems when running on Alpine Linux or other musl-based distros because phantomjs doesn't work.

By allow to disable this feature by a config value, the alpine build of https://github.com/hackmdio/docker-hackmd/pull/14 will no longer break.

geekyd:
Adds a config variable for enabling & disabling PDF exports

Fixes #561

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -384,7 +384,12 @@ function noteActions (req, res, next) {
         actionInfo(req, res, note)
         break
       case 'pdf':
-        actionPDF(req, res, note)
+        if (config.allowpdfexport) {
+          actionPDF(req, res, note)
+        } else {
+          logger.error('PDF export failed: Disabled by config. Set "allowpdfexport: true" to enable. Check the documentation for details')
+          response.errorForbidden(res)
+        }
         break
       case 'gist':
         actionGist(req, res, note)
--- a/lib/response.js
+++ b/lib/response.js
@@ -69,6 +69,7 @@ function showIndex (req, res, next) {
     ldap: config.isLDAPEnable,
     email: config.isEmailEnable,
     allowemailregister: config.allowemailregister,
+    allowpdfexport: config.allowpdfexport,
     signin: req.isAuthenticated(),
     infoMessage: req.flash('info'),
     errorMessage: req.flash('error')
@@ -98,7 +99,8 @@ function responseHackMD (res, note) {
     google: config.isGoogleEnable,
     ldap: config.isLDAPEnable,
     email: config.isEmailEnable,
-    allowemailregister: config.allowemailregister
+    allowemailregister: config.allowemailregister,
+    allowpdfexport: config.allowpdfexport
   })
 }
 
--- a/README.md
+++ b/README.md
@@ -145,9 +145,10 @@ Environment variables (will overwrite other server configs)
 | HMD_LDAP_SEARCHFILTER | `(uid={{username}})` | LDAP filter to search with |
 | HMD_LDAP_SEARCHATTRIBUTES | no example | LDAP attributes to search with |
 | HMD_LDAP_TLS_CA | `server-cert.pem, root.pem` | Root CA for LDAP TLS in PEM format (use comma to separate) |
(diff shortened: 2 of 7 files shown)


SISheogorath:
👍  for this PR!

Please have a look at:

https://github.com/hackmdio/hackmd/blob/5ce8f40eac7fccfcfca175cf79ccfa86a84bc113/lib/response.js#L285-L307

and

https://github.com/hackmdio/hackmd/blob/5ce8f40eac7fccfcfca175cf79ccfa86a84bc113/lib/response.js#L384-L386

We need a server-side implementation because this export function uses PhantomJS which doesn't work on [musl libc](https://www.musl-libc.org/), which is the library for alpine. To make a long story short, when you run the pdf export on the alpine based container, HackMD crashes (DoS attack vector) and we want to prevent this.

Can you please have a look at this with this PR, too?

geekyd:
@SISheogorath Thanks for the quick response. I will take a look at it. Just to be clear about the requirements here, we can go for something like this:

```
 function actionPDF (req, res, note) { 
   var body = note.content 
   var extracted = models.Note.extractMeta(body) 
   var title = models.Note.decodeTitle(note.title) 
  
   if (!fs.existsSync(config.tmppath)) { 
     fs.mkdirSync(config.tmppath) 
   } 
   var path = config.tmppath + '/' + Date.now() + '.pdf' 

   if(config.allowPDFExport) {
   markdownpdf().from.string(extracted.markdown).to(path, function () { 
     var stream = fs.createReadStream(path) 
     var filename = title 
     // Be careful of special characters 
     filename = encodeURIComponent(filename) 
     // Ideally this should strip them 
     res.setHeader('Content-disposition', 'attachment; filename="' + filename + '.pdf"') 
     res.setHeader('Cache-Control', 'private') 
     res.setHeader('Content-Type', 'application/pdf; charset=UTF-8') 
     res.setHeader('X-Robots-Tag', 'noindex, nofollow') // prevent crawling 
     stream.pipe(res) 
     fs.unlink(path) 
   }) 
  }
 } 
```

OR 

```
case 'pdf': 
   if(config.allowPDFExport) {
    actionPDF(req, res, note) 
  }
   break 
```
Please correct me if I'm wrong

SISheogorath:
After thinking about it, the last method looks better to me. but we should handle it correctly and return 403 as error code when the function is disabled.

SISheogorath:
Also please sign https://github.com/hackmdio/hackmd/pull/578 so we can make sure this and future changes are compatible with the new AGPL-3.0 license.

It would help us a lot to keep HackMD's CE awesome ;)

geekyd:
@SISheogorath I have signed the license change ([here](https://github.com/hackmdio/hackmd/pull/578#issuecomment-336739337))

lib/response.js
```diff
@@ -382,7 +384,11 @@ function noteActions (req, res, next) {
         actionInfo(req, res, note)
         break
       case 'pdf':
-        actionPDF(req, res, note)
+        if (config.allowPDFExport) {
+          actionPDF(req, res, note)
+        } else {
+          response.errorForbidden(res)
```
SISheogorath: We should also add `logger.error('PDF export failed: Disabled by config. Set "allowPDFExport: true" to enable. Check the documentation for details')` so no one is confused and we provide a helpful error message.

geekyd: @SISheogorath  Thanks for the suggestion. Added the error message

geekyd:
@SISheogorath Updated! Sorry for the delay.

SISheogorath:
@geekyd no problem. But please check your latest changes. -> https://github.com/hackmdio/hackmd/pull/589/files#diff-2ec1599c37878db948302f7e8152abf4L35

With this PR you would undo https://github.com/hackmdio/hackmd/pull/595

geekyd:
@SISheogorath Updated. :)

SISheogorath:
Thanks a lot!

PR #591, opened 2017-10-17
support Simplified Chinese and rename original zh to Traditional Chinese

Rwing:


SISheogorath:
Please fix the code to follow our code standard.

Run `npm test` to verify your changes locally. (Make aure you installed the `devDepensencies` in before using `npm install --dev`)

Rwing:
okay, it's done, sorry for my mistake

public/js/locale.js
```diff
@@ -4,6 +4,9 @@
 var lang = 'en'
 var userLang = navigator.language || navigator.userLanguage
 var userLangCode = userLang.split('-')[0]
+if (userLangCode === 'zh') {
+  userLangCode = userLang
+}
```
SISheogorath: I'm not sure why you add this code here. Can you please explain this change?

Rwing:
But after my tests, all browsers(ie, chrome, firefox) are return `zh-CN ` not `zh-Hans`

SISheogorath:
You are completely correct. I checked some other implementations and the `zh-hans` solution is a real corner case. The way you provided is usual. So 👍 for this.

We only need to make sure that the people who currently have already a language set (as a cookie) don't end up in a not working situation. Because right now this cookie would request `zh` as language which no longer exists.

Please add a fallback code to `zh-TW`, if (`zh`) is used, here:
https://github.com/Rwing/hackmd/blob/a03a6d97ffcb02602db35cb084b0bd365479f807/public/js/locale.js#L16

public/js/locale.js
Rwing: Sorry, I have re-read too, maybe this code is redundant because had https://github.com/Rwing/hackmd/blob/a03a6d97ffcb02602db35cb084b0bd365479f807/public/js/locale.js#L17

Rwing:
done. but a little ugly

a60814billy:
The translation seems good, it's not a big difference with traditional chinese language.

SISheogorath:
After playing around with the legacy local `zh` I realized that the migration code we added above is only used when people try to change their language.

To make sure that they don't need to reset their language, we do the following:

add a symlink `zh.json` to `zh-TW.json` but don't add anything further to the menu. On the server side, you should add `'zh-TW'` and `'zh-CN'` to the array but leave `'zh'`in there, too.

This way, it automatically uses the old `'zh'` translation for old users and everyone who sets the language newly will automatically use `zh-TW` and `zh-CN`.

Rwing:
Sorry, I dont know how to add a symlink zh.json to zh-TW.json can you help me? and is this way too complex?

ccoenen:
Just a thought: symlinks won't work if HackMD *server* is run on a windows machine. Granted this is probably only a tiny subset of installations.
(you need admin privileges to create a junction (symlink equivalent) on windows, which git usually does not have while cloning. Windows users will end up with a file named like the symlink, containing the target of the symlink as plain text.)

SISheogorath:
Creating symlinks:

On Linux and MacOS:
```console
ln -s locales/zh-TW.json locales/zh.json
```

On Windows open the git shell/git bash to run these commands.

---

About Windows and symlinks: It's not such a big problem because git-on-Windows will simply copy the file in this place instead of creating a link.

But in general, I'm not sure if we support windows for server-side installations. 🤔

SISheogorath:
Can you please squash the commits? Everything else is already fine ^^

How to do this:

```console
git rebase -i HEAD~4
git push -f
```

Replace the `pick` for all the later commands and replace it with a `fix` or `squash` to squash them into one big commit.

This helps us to keep the commit history clean.

Further information:
- http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html
- https://www.devroom.io/2011/07/05/git-squash-your-latests-commits-into-one/

Rwing:
Done, thank you very much for your help.

SISheogorath:
Thanks a lot for you work! ❤️

SISheogorath:
Oh can you please sign the #578 for the upcomming license change?

Issue #577, opened 2017-10-11
Empty export section should be hidden


SISheogorath:
When you run HackMD without external auth providers the export section exists but is empty. We shouldn't show an empty section.

![image](https://user-images.githubusercontent.com/8719867/31442135-b312409a-ae95-11e7-8ad6-2a18630c2e03.png)

This file should be the place to look for it: https://github.com/hackmdio/hackmd/blob/master/public/views/hackmd/header.ejs

kunal12422:
@SISheogorath can I work on this?

SISheogorath:
Sure :) Pull Request are welcome :heart: 

Feel free to ask if you have questions :smile:

geekyd:
@kunal12422 Have you started working on this??

PR #595, opened 2017-10-17
Hides empty export section

geekyd:
Fixes : #577 

Hides export section from header if no auth is present


SISheogorath:
Thanks a lot! :tada:

Issue #594, opened 2017-10-17
Remove remaining CSP issues


SISheogorath:
Since #585 we are able to configure the HSTS policy provided by [`helmet`](https://helmetjs.github.io/).

Let's take the next step to modern web security and add a [Content-Security-Policy](https://en.wikipedia.org/wiki/Content_Security_Policy).

This can be done by `helmet`, too. Details about how to do this can be found here: 
https://helmetjs.github.io/docs/csp/

Keep in mind we should allow people themselves to enable and extend them. We want to ship secure defaults, so they should be enabled by default.

Also, keep in mind that we have the `usecdn` option which will include external sources. These should be added when `usecdn` is enabled, while it shouldn't be allowed to use them by default when `usecdn` is disabled.

Last but not least there is some validation work to do. Make sure that the [features page](https://github.com/hackmdio/hackmd/blob/master/public/docs/features.md) is successfully rendered by default. Including all external embeddings like YouTube, Disqus, Google Analytics and more.

Pull Requests are very welcome ❤️ 

This is maybe interesting for someone who wants to work on web security and learn about the current technologies.

Some background information:
- [Introduction to CSP](https://scotthelme.co.uk/content-security-policy-an-introduction/)
- [Secure-headers.io for policy validation](https://securityheaders.io)
- [A little collection of knowledge about CSPs](https://github.com/ScottHelme/CSP-useful)

xxyy:
Since I also did #585, I'd like to do this one too. I figure that this will probably turn out to take some more time, since it also includes testing all features, plus figuring out all the external services.

Will we allow all images? I figure if we are only going to allow images from the configured provider, users will not be able to specify custom URLs, which doesn't seem like nice UX. (Of course a whitelist could be configurable if deemed a valid use case)

Also, is the feature list at `/feature` complete? (meaning if all issues there and the login/register form are fixed, there won't be any issues with variations?)

xxyy:
Update, looks like there is a significant number of inline styles in the views:
````
➜  views git:(master) ✗ grep -i --color -r "style=" . | wc -l
81
````
I'll be cleaning these up first in a separate PR, I guess. Using `unsafe-inline` for these relatively few doesn't seem like a viable 'solution' to me.

In unrelated news, `helmet` didn't promise too much, sending the headers is really straightforward.

SISheogorath:
**Awesome! ❤️ I love to hear/read this!**

[...]

PR #598, opened 2017-10-22
Implement basic CSP support

xxyy:
### Synopsis
In response to #594, this PR adds *basic* support for Content-Security-Policy headers. By default, some headers are sent in order to improve security for everyone. The default headers have been tuned for maximum usability, i.e. trying not to break anything. If they still do, these defects will be resolved as users open issues for them.

 ### Implementation overview
This PR utilises the Helmet library for CSP support. Due to the relative complexity of CSP directive computation, this logic has been placed in a separate file from `app.js`, `lib/csp.js`. For computation of the directives, first manually configured directives are added (`config.json`), then default directives, and finally, (if `usecdn`) special CDN directives. Custom directives are configured via a JSON object, so that even directives not used by default but supported by Helmet may be used. For complex custom setups, the default values can be disabled completely. By default, `upgrade-insecure-requests` is only specified if `usessl` is enabled (pointless otherwise). This can also be configured however.

### Implementation details
In the slide view, there is a dynamic inline script. For this script, a random nonce is generated and interpolated into that `<script>` tag by the server. The nonce is generated using the `uuid` library.

Also, the slide view's speaker notes depend entirely on a huge inline script. Due to that template being provided by the `reveal.js` library, I've chosen to include its SHA-256 hash in the CSP header. That way, we can allow that script without modifying that template. If `unsafe-inline` is enabled via the configuration file, neither of these two are added to `script-src` because they would nullify `unsafe-inline`.

Further, I've moved some inline configuration for the *MathJax* library that's common to multiple views into an extra `.js` file. That way the configuration won't be blocked by CSP. I've also removed some unnecessary JSON quotes in `README.md` and `config.json.example`.

### Known caveats
 * As discussed on [Gitter](https://gitter.im/hackmdio/hackmd?at=59e864385c40c1ba79bce664), we currently have to allow `unsafe-eval` because a lot of libraries are included via the webpack `script-loader` plugin, which only supports `eval()`. The solution for this issue is probably going to be different for each library and therefore is widely out of scope for this PR. Before realising this, I migrated `js-url` to be eval-free.
* The solution for the slide view's speaker view isn't clean at all and needs to be updated every time `reveal.js` change their speaker view JS.
* As discussed in #594, we're still allowing inline styles here. This is however a releatively small security risk compared to the other caveats.
* `connect-src` allows everything. This is because of the integrations. I've only been able to test the Gist import, as I'm missing API keys for everything else. I figured that the other integrations will also connect to different servers, and keeping track of them all might get out of hand.
* `object-src` allows everything. This is sadly necessary for the PDF viewer, since Chrome treats PDFs as objects. This makes sense, but is still inconvenient: This way, we're also allowing other plugins such as Flash and Java Applets.
* Since I'm not familiar with all of the project's features, it's entirely possible that I missed to test some.

### Further steps required for a tighter CSP policy
* Pursue [reveal.js](https://github.com/hakimel/reveal.js/issues) to remove the inline script from the speakers' view (they call it "notes plugin")
* Migrate library usage away from `eval()`, possibly using libraries with better module support, or finding a working [combination of loaders that emulates module support](https://webpack.github.io/docs/shimming-modules.html).
* Reduce the amount of inline styles in the views (This is mostly pointless due to some libraries dynamically adding inline styles via JS anyways)

A running version of this PR can (temporarily) be seen at https://me.l1t.li:1337/ . I can squash some commits to reduce the history footprint of this PR if desired.

Looking forward to hear what you think! :)


SISheogorath:
I tested this on a local machine (without HTTPS).

Worked well on Chromium, but failed on Firefox. Sadly without a useful error message x.x

Needs more investigation.

Firefox version 56.0 on Fedora 26

xxyy:
Weird, I'll be taking a look soon-ish.

xxyy:
Seems to be related to the migration of the `js-url` library to an eval-less loader. Since that's not necessary now, I reverted that commit - seems to have resolved the issue.

SISheogorath:
I want to move this to the first release after 1.0 because it needs more testing and I'm a bit short in time right now

SISheogorath:
Just checked the PR again to make it happen. Noticed that I still get errors. After some investigation it turns out that the GitHub button on the cover/index page is the source for this. Removed the iframe, no errors.

I just tried to embed a new one using https://github.com/ntkme/github-buttons which should be fine for CSP complaint ones, but didn't work perfect for me. After wasting around 20 minutes I wonder if we can't just put a link there without fancy buttons ._.

Opinions?

jackycute:
@SISheogorath True, place a link is faster and more secure.

SISheogorath:
@xxyy Thank you so much for your work! It's a pleasure to work with you!

You helped to secure HackMD a lot!

PR #606, opened 2017-10-29
Add Mattermost authentication strategy

DoubleMalt:
I added the possibility to authenticate against a Mattermost server. I was not sure about the version numbers, but I'm happy to change them to what they should be.


SISheogorath:
Awesome PR! Instantly coming with all needed changes. Very nice!

But as you see the CI process fails because of not following the coding standard. Can you fix this, please? Only minor issues ^^

You can test it yourself on your dev machine using `npm test`

Thanks a lot~

DoubleMalt:
Done.

SISheogorath:
Could you squash the commits using `git rebase -i HEAD~6` so we don't have to many "Fix" commits in our git history?

Also one important question: Is this new auth provider tested? Because I'm not really sure how to test it, since I don't run a mattermost instance myself.

DoubleMalt:
Squashed.

The new authentication mechanism is used at the moment for https://notes.doublemalt.net. I can create a test team there so someone can test it against it if desired.

Agreed to the license change.

SISheogorath:
When you run it (in production), that's fine for me :smile: 

Only want to prevent untested code.

We have to keep an eye on https://github.com/FortAwesome/Font-Awesome/issues/9727

Currently, there is no mattermost icon :/

But a missing icon will not prevent us from having this awesome feature!

Thanks a lot!

PR #609, opened 2017-10-30
Correcting grammatical errors related to the document char count tooltip

monoxane:
The tooltips didn't particularly make sense witht their wording before.


SISheogorath:
Please notice that there is a translation happening using the `./locales/`. Please make sure it's not already covered there.

PR #619, opened 2017-11-09
WIP: ability to use asciidoc

syzer:


ccoenen:
Hi! This is a very major change, in my opinion. Has this been discussed previously?

ccoenen:
For now I assume this has been opened by accident. I assume this was meant to be a PR for hackadoc, the asciidoc-fork of hackmd. It this was indeed supposed to be for hackmd, feel free to reach out so we can talk about it!

syzer:
@ccoenen you are right.
The idea was to have a tag that will change to use markdown or asciidoc parser
But that would would be a big change : in the parser(that he could load different parsers based on node type)
So parser would do (AST => compliled node) mapping
based on the document type.

SISheogorath:
We could actually add a flag in the DB about ASCII or Markdown and than we only need to replace a few views.

Instead of replacing the parser on the fly we would add the possibility to create ASCIIDoc and Markdown document on the "Create note" button and would solve the parser problem easily. But it needs some work and creates a way bigger dependency tree.

How about this way?

syzer:
Maybe like that:
you could take a look of issues that are not entirely working:
https://github.com/hackergarten/hackadoc/issues

and tell me what are important features to get fixed,
so we could merge to master

PR #627, opened 2017-11-24
Prevent XSS vulnerbility by srcdoc in iframe

### LINKED ISSUE #626: XSS Vulnerability in Markdown Editor
Hello, 

I found a XSS vulnerability in markdown editor. As a proof of concept, you can use payload like this to trigger a XSS:
```
<iframe srcdoc="<script>alert(document.domain)</script>">
```

The relevant code may be here:
https://github.com/hackmdio/hackmd/blob/master/public/js/render.js#L20-L21
```
// allow ifram tag with some safe attributes
whiteList['iframe'] = ['allowfullscreen', 'name', 'referrerpolicy', 'sandbox', 'src', 'srcdoc', 'width', 'height']
```

About `srcdoc` of `iframe`, the explanation of the MDN web docs is 
( https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe ) :

> The content of the page that the embedded context is to contain. This attribute is expected to generally be used together with the `sandbox` attribute. If a browser supports the `srcdoc` attribute, it will override the content specified in the src attribute (if present). If a browser does NOT support the `srcdoc` attribute, it will show the file specified in the src attribute instead (if present). Note that if the content of the attribute contains a script tag then a closing script tag is required for the script to run, even if nothing else comes after the script.

So as you can see, you can use `srcdoc` of `iframe` to execute JavaScript, and you will find the context domain is the same as parent's.

SISheogorath:
Fixes #626 

I don't think there are many people using this option, so it can be safely removed.

### DIFF
--- a/public/js/render.js
+++ b/public/js/render.js
@@ -18,7 +18,7 @@ whiteList['style'] = []
 // allow kbd tag
 whiteList['kbd'] = []
 // allow ifram tag with some safe attributes
-whiteList['iframe'] = ['allowfullscreen', 'name', 'referrerpolicy', 'sandbox', 'src', 'srcdoc', 'width', 'height']
+whiteList['iframe'] = ['allowfullscreen', 'name', 'referrerpolicy', 'sandbox', 'src', 'width', 'height']
 // allow summary tag
 whiteList['summary'] = []
 


ccoenen:
I believe the benefit greatly outweighs the cost. If there's someone using this option, I'm really sorry for them, but this is plainly dangerous.

PR #633, opened 2017-11-28
Support SAML authentication

nakaeeee:
I added SAML authentication using [bergie/passport-saml](https://github.com/bergie/passport-saml) library.
SAML auth is common way for enterprise use as well as LDAP.
Typical SAML ID-provider services include ADFS, OpenAM, OneLogin, AzureAD and so on.

I tested it with OpenAM with generic parameter settings.
User creation was made with reference to [lib/web/auth/ldap/index.js](https://github.com/hackmdio/hackmd/blob/master/lib/web/auth/ldap/index.js#L26-L64).

I would appreciate if you could accept :blush:


SISheogorath:
Thanks for this PR! I already thought about implementing this 👍 

Just a few questions:

Why is there a configurable callback URL for SAML? https://github.com/hackmdio/hackmd/pull/633/files#diff-12ef3ed9fd13e4ccf3bd7dc3701a5240R82

Also is there any reason why we can't replicate the SAML config to environment variables?

And as to do: It would be lovely to provide a short description of the config options in the README and maybe add some [more detailed guide](https://github.com/hackmdio/hackmd/blob/master/docs/guides/auth.md) into the `./docs/guides/` directory. This would help new people to adopt this and also me, and some others to test this setup 😄

nakaeeee:
Thanks for your review!

In the case of via proxy, I was worried that the URL will be different between outside and inside. I have just checked it and found that there is no problem if setting `domain`.
I'm going to fix the callback URL to unconfigurable. Thanks!

I was wondering about how to use `config.json` and environment variables properly for different purposes, and I guessed that environment variables might be due to reason that does not want to write in the file like secret key. Therefore, narrowed down the items.
Total of about 12 items will be added, but if that's OK then I will add it.

Agreed to the guide and short description.
I will describe for setting to OneLogin or AzureAD which is easy to prepare IDP as a case example. Please give me a little more time.

nakaeeee:
Modified the specification and added the guide. Could you please review again? :bowing_man:

nakaeeee:
I signed it. I'm happy to be of use to you. :smile:

Issue #634, opened 2017-11-28
sslcapath bug


laysdra7265:
I might encounter a bug with ssl config.
The config in document said that it was an array 
`sslcapath:['./cert/COMODORSAAddTrustCA.crt']`
The problem is `lib/config/index.js:98:25`it try to join it as a path and throw a error.
```console
TypeError: Path must be a string. Received [ 'ssl/ca_bundle.crt' ]
    at assertPath (path.js:7:11)
    at Object.join (path.js:1218:7)
    at Object.<anonymous> (/home/hackmd/lib/config/index.js:98:25)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)

```
If I try to set sslcapath as string path,than app.js can't handle it correctly.
``` console
[...]

PR #636, opened 2017-11-28
Fix sslcapath bug

laysdra7265:
#634 Thanks for your suggestion I finally split two pull request.
@SISheogorath

### DIFF
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -96,7 +96,7 @@ config.isPDFExportEnable = config.allowpdfexport
 
 // generate correct path
 config.sslcapath.forEach(function (capath, i, array) {
-    array[i] = path.resolve(appRootPath, capath)
+  array[i] = path.resolve(appRootPath, capath)
 })
 
 config.sslcertpath = path.join(appRootPath, config.sslcertpath)
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -95,14 +95,9 @@ config.isLDAPEnable = config.ldap.url
 config.isPDFExportEnable = config.allowpdfexport
 
 // generate correct path
-config.sslcapath = (function () {
-  var i, len, results
-  results = []
-  for (i = 0, len = config.sslcapath.length; i < len; i++) {
-    results.push(path.resolve(appRootPath, config.sslcapath[i]))
-  }
-  return results
-})()
+config.sslcapath.forEach(function (capath, i, array) {
+    array[i] = path.resolve(appRootPath, capath)
+})
 
 config.sslcertpath = path.join(appRootPath, config.sslcertpath)
 config.sslkeypath = path.join(appRootPath, config.sslkeypath)
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -95,7 +95,15 @@ config.isLDAPEnable = config.ldap.url
 config.isPDFExportEnable = config.allowpdfexport
 
 // generate correct path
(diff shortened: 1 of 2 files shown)


SISheogorath:
Well, I said moving the code, so we can remove the validation in the app.js since we already do the config validation in the config module.

No need to validate it twice.

laysdra7265:
But what the code does is not validation ,it is reading each ca file from path.
https://github.com/hackmdio/hackmd/blob/c794412714cdd7b0f6ebfadaf108f72098004d64/app.js#L49-L56

SISheogorath:
Yes, you are right, I misread it. Thanks for pointing me to this ^^

lib/config/index.js
```diff
@@ -95,7 +95,15 @@ config.isLDAPEnable = config.ldap.url
 config.isPDFExportEnable = config.allowpdfexport
 
 // generate correct path
-config.sslcapath = path.join(appRootPath, config.sslcapath)
+config.sslcapath = (function () {
+  var i, len, results
+  results = []
+  for (i = 0, len = config.sslcapath.length; i < len; i++) {
+    results.push(path.join(appRootPath, config.sslcapath[i]))
```
SISheogorath: Please use [`path.resolve()`](https://nodejs.org/api/path.html#path_path_resolve_paths) instead of path.join.

laysdra7265: Why don't we replace other path.join() in index.js into path.resolve()?

SISheogorath: We'll do this while refactoring, but right now we touch this part of the code and we should use `resolve()` since it doesn't break absolute paths. :)

laysdra7265: Why we need to use resolve() instead of join()?As I know they seems do the same thing there.

SISheogorath: While join will simply append the added path, resolve does always return an absolute path to the file. And it can handle absolute path as input, which join doesn't.

so when you put `path.join("/opt/hackmd", "/tmp/somefile")` you'll get `/opt/hackmd/tmp/somefile` while `path.resolve("/opt/hackmd", "/tmp/somefile")` returns `/tmp/somefile`.

laysdra7265: oh well.I got it.I just wondered why we'll need put two absolute path as parameter,that really weird since it isn't necessary.
Because if I specified an absolute path in config,the join() can't handle it as expected,right?:)

SISheogorath: `resolve` takes absolute paths as well as relatives, but always returns an absolute one.

As mentioned before: `path.resolve("/opt/hackmd", "/tmp/somefile")` returns `/tmp/somefile`.
But also possible: `path.resolve("/opt/hackmd", "config/somefile")` returns `/opt/hackmd/config/somefile`.

So `resolve` does both, while `join` can only handle relative paths.

SISheogorath:
Can you please rebase this to the latest master?

---

## How to do this?

1. Check what remotes exist: `git remote -v`
2. If upstream doesn't exist, add it: `git remote add upstream https://github.com/hackmdio/hackmd.git`
3. Stash your WIP: `git stash`
4. Checkout master: `git checkout master`
5. Pull latest changes from upstream: `git pull upstream master`
6. Checkout the branch you worked on before: `git checkout <your branch>`
7. Rebase to new master branch: `git rebase master`
8. Resolve conflicts if they appear
9. Get your changes back: `git stash pop`

laysdra7265:
wow,thanks for the tips,I don't know I can do this with git. XD

PR #643, opened 2017-12-02
Add support for qq avatar for qq email account

laysdra7265:
Use qq avatar api when user is from @qq.com.

### DIFF
--- a/lib/models/user.js
+++ b/lib/models/user.js
@@ -147,11 +147,20 @@ module.exports = function (sequelize, DataTypes) {
         return photo
       },
       parseProfileByEmail: function (email) {
-        var photoUrl = 'https://www.gravatar.com/avatar/' + md5(email)
-        return {
-          name: email.substring(0, email.lastIndexOf('@')),
-          photo: photoUrl + '?s=96',
-          biggerphoto: photoUrl + '?s=400'
+        if (email.toLocaleLowerCase().indexOf('@qq.com') > 0) {
+          let photoUrl = 'http://q1.qlogo.cn/g?b=qq&nk=' + email.substring(0, email.lastIndexOf('@'))
+          return {
+            name: email.substring(0, email.lastIndexOf('@')),
+            photo: photoUrl + '&s=100',
+            biggerphoto: photoUrl + '&s=640'
+          }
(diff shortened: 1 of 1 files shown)


SISheogorath:
The basic idea is nice, but from a privacy perspective, it's a nogo.

E-Mail addresses are sensitive data. Not so much in HackMD, but in general. And this integration would unmask user's entire email address.

How? -> Since the first part of the email address is part of the Image URL and the image URL is specified for a domain, you get user@domain which is the mail address.

As long as they don't mask the e-mail address like gravatar or even better, I don't think we want to merge this, sorry.

ccoenen:
I fully support @SISheogorath's reasoning. Even if this were behind a opt-in server-wide setting, it wouldn't be imeediately obvious for that server's users. If this were a per-user setting (or perhaps even better: if one would just proxy the image from qq) this problem would go away.

laysdra7265:
I'll rewrite with a encrypted way later.

SISheogorath:
Can you please rebase this to the latest master?

---

## How to do this?

1. Check what remotes exist: `git remote -v`
2. If upstream doesn't exist, add it: `git remote add upstream https://github.com/hackmdio/hackmd.git`
3. Stash your WIP: `git stash`
4. Checkout master: `git checkout master`
5. Pull latest changes from upstream: `git pull upstream master`
6. Checkout the branch you worked on before: `git checkout <your branch>`
7. Rebase to new master branch: `git rebase master`
8. Resolve conflicts if they appear
9. Get your changes back: `git stash pop`

Rwing:
I think it's not good idea for privacy. 
First, qq.com not allow invoke its avator image url in public document, there isn't open api like as gravator.
Second, users maybe not want display his qq.com avator in this site

laysdra7265:
@Rwing or show qq avatar when user is login via qq auth?

Rwing:
yeah, maybe 'qq oauth' is a better solution, there is a document of open api http://wiki.connect.qq.com/get_user_info

laysdra7265:
Well ok,I'll just use it as my own feature on my server :)
And pull another PR about 'qq oauth' later.

PR #650, opened 2017-12-09
Add setting `ldap.usernameField`

LukasKalbertodt:
This determines which ldap field is used as the username on
HackMD. By default, the "id" is used as username, too. The id
is taken from the fields `uidNumber`, `uid` or
`sAMAccountName`. To give the user more flexibility, they can
now choose the field used for the username instead.

---

Context: I want to use HackMD in an LDAP environment, where the ldap field chosen for the user id is just a number. Seeing a number on the top right in HackMD is a bit ugly. So with this PR, I can choose which ldap field is used as username.

I hope this PR is fine. I don't have any experience in JS. I tested this patch with my ldap server and everything runs fine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,6 +170,7 @@ There are some configs you need to change in the files below
 | HMD_LDAP_SEARCHBASE | `o=users,dc=example,dc=com` | LDAP directory to begin search from |
 | HMD_LDAP_SEARCHFILTER | `(uid={{username}})` | LDAP filter to search with |
 | HMD_LDAP_SEARCHATTRIBUTES | no example | LDAP attributes to search with |
+| HMD_LDAP_USERNAMEFIELD | `uid` | The LDAP field which is used as the username on HackMD |
 | HMD_LDAP_TLS_CA | `server-cert.pem, root.pem` | Root CA for LDAP TLS in PEM format (use comma to separate) |
 | HMD_LDAP_PROVIDERNAME | `My institution` | Optional name to be displayed at login form indicating the LDAP provider |
 | HMD_SAML_IDPSSOURL | `https://idp.example.com/sso` | authentication endpoint of IdP. for details, see [guide](docs/guides/auth.md#saml-onelogin). |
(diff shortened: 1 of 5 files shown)


SISheogorath:
Thanks for your contribution! 👍 

And if you don't know JS, even nicer to see you made it!

Just a few questions, because it looks very complete:

First of all, can you tell what type of LDAP you have tested it with? OpenLDAP? Active Directory? Others? Because we should at least verify that OpenLDAP and Active Directory work.

Also could you please add the description for it in the other config section: https://github.com/hackmdio/hackmd#application-settings-configjson

I know it's not perfect right now with your environment configs, but currently go this way and it sadly leads to some dublicates.

LukasKalbertodt:
Thank you for your time reviewing this PR!

> Also could you please add the description for it in the other config section: https://github.com/hackmdio/hackmd#application-settings-configjson

I don't see any mention of any of the other ldap settings in the `config.json` section of the README. Should I add a complete description of all ldap parameters? Or am I missing something?

> First of all, can you tell what type of LDAP you have tested it with? OpenLDAP? Active Directory? Others? Because we should at least verify that OpenLDAP and Active Directory work.

Honestly, I don't know and I have no clue how to find out right now. I can ask some administrators on monday, but I don't think I will find out during the weekend still. Anyway, I only have this one ldap server to test it with, so I cannot test it for both.

Lastly, I really don't think this change is ldap server dependent. It's basically just how HackMD handles the data it gets from the server. (Not to say testing is useless, but yeah...)

SISheogorath:
> I don't see any mention of any of the other ldap settings in the config.json section of the README. Should I add a complete description of all ldap parameters? Or am I missing something?

I just noticed that we didn't put any auth providers into this section. So scratch that statement 😄 

> Lastly, I really don't think this change is ldap server dependent. It's basically just how HackMD handles the data it gets from the server.

I agree that it's not mandatory but I just want to make sure it works well. Maybe someone will make it to test it until next week, otherwise we will get this change in without more testing. And you don't need to test both, that's something that should happen in the community ;)

LukasKalbertodt:
@SISheogorath FYI: the system I tested it with doesn't run Active Directoy, so probably OpenLDAP.

SISheogorath:
Nice, thanks for the feedback

PR #654, opened 2017-12-11
Add fix for pinyin input method on Windows

SISheogorath:
@laysdra7265 can you test and validate this please?


jackycute:
Thanks for heads up @SISheogorath 
But according to the https://github.com/mizchi/codemirror-composition-mod
It seems like this PR lacking of import the mod and set the option for CodeMirror.
Could @@laysdra7265 give us help?

SISheogorath:
Seems to not be worked on… Closing for now.

PR #660, opened 2017-12-28
Updated yarn.lock.

edoardoo:
It should fix the yarn.lock problem.


SISheogorath:
I did this myself because I wasn't sure if you may have additional dependencies in your own fork. Thanks for your work!

PR #663, opened 2017-12-29
Fullscreen

edoardoo:
Hi, 
as you might have seen I'm working a bit on a personal_use/customized_version of HackMD. 
I've implemented a fullscreen feature which I've found relevant in a number of occasions, from presentation to distraction-free environment.

screen shot 2017-12-29 at 6 23 53 pm

I've placed the button to activate the fullscreen mode inside the menu. You can check it into [my heroku instance](https://wriwri.herokuapp.com/features).  
In any case it needs a code-review, at least to check if I did put the code in the right files. 
  
I hope this doesn't sound like a way to force features inside the app. It's just a feature that I wanted to implement myself anyway, if you find it appropriate I could work on it to effectively merge it.
  
Also, I'm terribly sorry about the unsigned code. Can I fix it somehow?  Should I just add a signed empty commit to the pull request?

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -102,5 +102,8 @@
 	"Select From Available Snippets": "Select From Available Snippets",
 	"OR": "OR",
 	"Export to Snippet": "Export to Snippet",
-	"Select Visibility Level": "Select Visibility Level"
+	"Select Visibility Level": "Select Visibility Level",
+	"Dark Theme": "Dark Theme",
+	"Night Theme": "Night Theme",
+	"Fullscreen": "Fullscreen"
 }
--- a/public/js/index.js
+++ b/public/js/index.js
@@ -419,6 +419,7 @@ Visibility.change(function (e, state) {
 $(document).ready(function () {
   idle.checkAway()
   checkResponsive()
+  addExitFullscreenListeners()
     // if in smaller screen, we don't need advanced scrollbar
   var scrollbarStyle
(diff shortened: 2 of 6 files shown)


locales/en.json
```diff
@@ -102,5 +102,8 @@
 	"Select From Available Snippets": "Select From Available Snippets",
 	"OR": "OR",
 	"Export to Snippet": "Export to Snippet",
-	"Select Visibility Level": "Select Visibility Level"
-}
+	"Select Visibility Level": "Select Visibility Level",
+	"Dark Theme": "Dark Theme",
+	"Night Theme": "Night Theme",
```
SISheogorath: These theme commits should be separated

edoardoo: Sorry, it splipped. Gonna fix it later.

edoardoo:
@SISheogorath I did put the button on the View selector side (where the night theme is), but changed it because I wanted to maintain the interface clean.   

Now that I'm testing it I can see that adding the app to home screen in Android Chrome gives the same result on mobile as F11 does on desktop.   
So the only use case remains when the user wants to use it fullscreen on a mobile device and quickly needs to switch view without adding it to home screen, aside of course of being a shortcut for desktop users.

At this point, as you said, I'm not even sure if it's worth to add code to be maintained.   
For me it's okay to drop it, let me know.

PR #673, opened 2018-01-11
Allow posting new note with content

fooker:
This allows to specify initial content while creating a new note.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -117,7 +117,8 @@ function newNote (req, res, next) {
   }
   models.Note.create({
     ownerId: owner,
-    alias: req.alias ? req.alias : null
+    alias: req.alias ? req.alias : null,
+    content: req.body ? req.body : ''
   }).then(function (note) {
     return res.redirect(config.serverurl + '/' + LZString.compressToBase64(note.id))
   }).catch(function (err) {
--- a/lib/web/noteRouter.js
+++ b/lib/web/noteRouter.js
@@ -4,10 +4,14 @@ const Router = require('express').Router
 
 const response = require('../response')
 
+const {markdownParser} = require('./utils')
+
 const noteRouter = module.exports = Router()
 
 // get new note
 noteRouter.get('/new', response.newNote)
+// post new note with content
+noteRouter.post('/new', markdownParser, response.newNote)
 // get publish note
 noteRouter.get('/s/:shortid', response.showPublishNote)
 // publish note actions
--- a/lib/web/utils.js
+++ b/lib/web/utils.js
@@ -7,3 +7,10 @@ exports.urlencodedParser = bodyParser.urlencoded({
   extended: false,
   limit: 1024 * 1024 * 10 // 10 mb
 })
+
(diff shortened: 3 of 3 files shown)


ccoenen:
Interesting! Can you give an example on when you use this? And perhaps an example of how you use it?

fooker:
We would like to use hackmd for filling protocols of regular meetings. Using this, we preseed the per-meeting note with a protocol template and publish it to the participants. We're planing to do this with an IRC bot, but it is a hacky shell script right now. It basicly does something like this:
```
curl -v -XPOST -H 'Content-Type: text/markdown' 'http://localhost:3000/new' -d @template.md
```

SISheogorath:
Nice work!

Could you please sign-off the commit?

https://github.com/hackmdio/hackmd/blob/master/CONTRIBUTING.md#sign-your-work

fooker:
Sure. Updated...

SISheogorath:
Needs some docs on the Features page.

Thanks for your contribution! 👍

rmst:
Hi is it possible to use this on hackmd.io ? I tried

```bash
curl -v -XPOST -H 'Content-Type: text/markdown' 'https://hackmd.io/new' -d test123
```
and got back
```html
<!DOCTYPE html>
<html lang="en">
<head>
<meta charset="utf-8">
<title>Error</title>
</head>
<body>
<pre>Cannot POST /new</pre>
</body>
</html>
```

SISheogorath:
@rmst No, this was after the split between HackMD EE/HackMD.io and CodiMD (former HackMD CE), so it never got there.

rmst:
Oh right, I've made that mistake before.. github.com/hackmdio just looks soo much like it belongs to hackmd.io.
So I found https://notes.ethereum.org and the curl worked, so I suppose that's the official site where CodiML is hosted? Is there a place where I could read up on the story of CodiMD and hackmd.io etc?

I have to say both are stunningly similar. The only functional differences I've noted so far are:
- hackmd.io has link-autocomplete
- notes.ethereum.org has this 'API' for inserting content into new notes.
Are there any other functional differences?

SISheogorath:
We have actually nothing to do with https://notes.ethereum.org/ besides the fact that they happen to run an instance of CodiMD

CodiMD has no official hosted instance. There is https://hackmd-ce.herokuapp.com which is our test/demo instance which runs the latest master.

We recommend to self-host your instance if you want to use CodiMD. We also have an upcoming community call on sunday, if you want to meet some community members (see #911).

About the HackMD vs CodiMD story: Please check #720 and the "TL;DR" in our readme: https://github.com/hackmdio/codimd#hackmd-ce-became-codimd

rmst:
Oh interesting! Thanks for giving me all that info despite this being so off topic!

patcon:
Fwiw, seems to not work on notes.ethereum.org anymore :man_shrugging:

SISheogorath:
@patcon seems like they setup a SaaS instance by the HackMD team. This doesn't support this feature.

PR #675, opened 2018-01-14
Add Azure Blob Storage support

ahoka:
Add support for storing images in an Azure Blob Storage container.

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,6 +212,8 @@ There are some config settings you need to change in the files below.
 | `HMD_MINIO_ENDPOINT` | `minio.example.org` | Address of your Minio endpoint/instance |
 | `HMD_MINIO_PORT` | `9000` | Port that is used for your Minio instance |
 | `HMD_MINIO_SECURE` | `true` | If set to `true` HTTPS is used for Minio |
+| `HMD_AZURE_CONNECTION_STRING` | no example | Azure Blob Storage connection string |
+| `HMD_AZURE_CONTAINER` | no example | Azure Blob Storage container name (automatically created if non existent) |
 | `HMD_HSTS_ENABLE` | ` true`  | set to enable [HSTS](https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security) if HTTPS is also enabled (default is ` true`) |
 | `HMD_HSTS_INCLUDE_SUBDOMAINS` | `true` | set to include subdomains in HSTS (default is `true`) |
 | `HMD_HSTS_MAX_AGE` | `31536000` | max duration in seconds to tell clients to keep HSTS status (default is a year) |
@@ -261,7 +263,7 @@ There are some config settings you need to change in the files below.
 | `documentMaxLength` | `100000` | note max length |
 | `email` | `true` or `false` | set to allow email signin |
 | `allowEmailRegister`  | `true` or `false` | set to allow email register (only applied when email is set, default is `true`. Note `bin/manage_users` might help you if registration is `false`.) |
-| `imageUploadType` | `imgur`(default), `s3`, `minio` or `filesystem` | Where to upload image
+| `imageUploadType` | `imgur`(default), `s3`, `minio`, `azure` or `filesystem` | Where to upload image
 | `minio` | `{ "accessKey": "YOUR_MINIO_ACCESS_KEY", "secretKey": "YOUR_MINIO_SECRET_KEY", "endpoint": "YOUR_MINIO_HOST", port: 9000, secure: true }` | When `imageUploadType` is set to `minio`, you need to set this key. Also checkout our [Minio Image Upload Guide](docs/guides/minio-image-upload.md) |
 | `s3` | `{ "accessKeyId": "YOUR_S3_ACCESS_KEY_ID", "secretAccessKey": "YOUR_S3_ACCESS_KEY", "region": "YOUR_S3_REGION" }` | When `imageuploadtype` be set to `s3`, you would also need to setup this key, check our [S3 Image Upload Guide](docs/guides/s3-image-upload.md) |
 | `s3bucket` | `YOUR_S3_BUCKET_NAME` | bucket name when `imageUploadType` is set to `s3` or `minio` |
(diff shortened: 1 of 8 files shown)


SISheogorath:
Anyone else using Azure for HackMD hosting? I don't and I don't want to sign up to validate this feature.

Maybe someone can help to test this feature.

ahoka:
You can use an emulator: https://github.com/arafato/azurite

ccoenen:
I'll formulate this somewhat provocatively, please don't take it personally: If nobody else is interested enough to even test it, it's probably not a good idea to merge it.

ahoka:
It takes more to offend me, I'm East European. However, I think your argument is flawed because:

- Most of your users are not present on GitHub
- Most of your users will only use official packages

Your company is using hackmd for internal docs, so we would be glad to get better Azure support. If this gets merged, we plan to contribute Azure AD authentication support in the future.

So please don't make us use SharePoint, that would be cruel!

SISheogorath:
> So please don't make us use SharePoint, that would be cruel!

That's an argument that gets me 😆 

Maybe I'll sign up for an azure account later and really verify it. It makes sense to me to support it, it's just that I'm not willing to manage millions of accounts for all kind of features just for verification and testing. And when we look at HackMD's feature set it's such a huge number of services that are supported, that it becomes hard to keep track of all.

I would really welcome some more community side verification and testing, that's all.

SISheogorath:
Can you provide some documentation about how to set it up and how to get the needed storage?

For example when I check the Microsoft page I'm a bit confused which of these storages I should use:
https://azure.microsoft.com/en-us/pricing/details/storage/

Maybe take #698 as an example.

SISheogorath:
Mhm just tried to sign up for Azure but I'm not really willed to share my credit card information with Microsoft .-.

So this way I can't verify it. Sorry.

SISheogorath:
@Kemyke well, while I welcome the gesture, it would probably be easier when you just verify the vranch to work.

The easiest way to do so is uncommenting the build instructions in the docker-compose.yml from docker-hackmd and point it to @ahoka's repository and branch, then run `docker-compose build` edit the configs (in the compose file), start it using `docker-compose up` and see if the image upload works and if it is persistent as expected.

When @ahoka rebased the changes to the current code base so we can merge it, and it works fine, I'm fine with merging it.

Kemyke:
@SISheogorath 

Here are my steps to test this feature:

- Create a blob in Azure
- Clone the docker-hackmd repository
- Edit docker-compose.yml as you recommended
```
# Using version 3 to provide play-with-docker badge
# You can change to version 2 without breaking.
#version: '2'
version: '3'
services:
  database:
    # Don't upgrade PostgreSQL by simply changing the version number
    # You need to migrate the Database to the new PostgreSQL version
    image: postgres:9.6-alpine
    environment:
      - POSTGRES_USER=hackmd
      - POSTGRES_PASSWORD=hackmdpass
      - POSTGRES_DB=hackmd
    volumes:
      - database:/var/lib/postgresql/data
    networks:
      backend:
    restart: always

  # MySQL example
  # Most of the documentation that applies to PostgreSQL applies also to MySQL
  #database:
  #    # You should be able to upgrade MySQL without problems
  #    # but to make sure no even when a problem appears you
  #    # should have a backup
  #    image: mariadb:10
  #    environment:
  #      - MYSQL_USER=hackmd
  #      - MYSQL_PASSWORD=hackmdpass
  #      - MYSQL_DATABASE=hackmd
  #      - MYSQL_ALLOW_EMPTY_PASSWORD=true
  #    volumes:
  #      - database:/var/lib/mysql
  #      # This config provides UTF-8 support to the database by default
  #      # If this config is not used, HackMD breaks as it tries to write
  #      # UTF-8 to a latin database.
  #      - ./resources/utf8.cnf:/etc/mysql/conf.d/utf8.cnf
  #    networks:
  #      backend:
  #    restart: always

  app:
    # Uncomment the following section to build the image yourself:
    build:
      context: .
      dockerfile: debian/Dockerfile
      args:
        - "VERSION=master"
        - "HACKMD_REPOSITORY=https://github.com/ahoka/hackmd.git"
    #image: hackmdio/hackmd:1.1.0-ce
    environment:
      # DB_URL is formatted like: <databasetype>://<username>:<password>@<hostname>/<database>
      # Other examples are:
      # - mysql://hackmd:hackmdpassdatabase:3306/database
      # - sqlite:///data/sqlite.db (NOT RECOMMENDED)
      # - For details see the official sequelize docs: http://docs.sequelizejs.com/en/v3/
      - HMD_IMAGE_UPLOAD_TYPE=azure
      - HMD_DB_URL=postgres://hackmd:hackmdpass@database:5432/hackmd
      - HMD_AZURE_CONNECTION_STRING=DefaultEndpointsProtocol=https;AccountName=hackmdbackup;AccountKey=<MyAccountKey>;EndpointSuffix=core.windows.net
      - HMD_AZURE_CONTAINER=content
    ports:
      # Ports that are published to the outside.
      # The latter port is the port inside the container. It should always stay on 3000
      # If you only specify a port it'll published on all interfaces. If you want to use a
      # local reverse proxy, you may want to listen on 127.0.0.1.
      # Example:
      # - "127.0.0.1:3000:3000"
      - "3000:3000"
    networks:
      backend:
    restart: always
    depends_on:
      - database

# Define networks to allow best isolation
networks:
  # Internal network for communication with PostgreSQL/MySQL
  backend:

# Define named volumes so data stays in place
volumes:
  # Volume for PostgreSQL/MySQL database
  database:
```
- Build and up the compose file
- Upload a file in a guest note

I could sucessfully upload an image to the Azure blob. You can access my test [instance](http://hackmdazureblobtest.westeurope.cloudapp.azure.com:3000).

Here you can see the result in the UI:


And as you can see there is the image in the blob:


I also tested that the HackMD creates the blob container automatically, you don't need to create it by yourself.

SISheogorath:
Very nice! Can you sign-off your commits? Then I'll merge it :)

To prevent you from redoing all the commits, you can just use `git rebase --signoff`.

Sign-off is required by our [contribution guideline](https://github.com/hackmdio/hackmd/blob/master/CONTRIBUTING.md#sign-your-work).

ahoka:
Hi guys, I have updated and signed-off the commit.

SISheogorath:
Thanks a lot for your work @ahoka and also thanks for your patience. Great job!

ahoka:
Thanks for including my changes. I have opened a small PR with a fix for a minor error that I have found during updating the branch. Would you also accept a patch with Azure AD support (Azure's OAUTH service, unrelated to Active Directoy)?

SISheogorath:
Can you check if #784 would already work for Azure?

Would be a great test as well ^^ Then we don't need to add more code, just some docs.

Kemyke:
@SISheogorath Is it possible to include this feature in the 1.2.0-CE release?

PR #678, opened 2018-01-16
Fix anchor id to keep uppercase characters

jackycute:
id shouldn’t be converted to lowercase since id attribute is case sensitive


SISheogorath:
Have you verified this doesn't break anything or need an migration?

jackycute:
@SISheogorath 
It's all good, I've done some manual checks.
Also, this only affects the markdown parser so don't worry about the schema.

SISheogorath:
Ah, perfect, thanks!

PR #680, opened 2018-01-18
Update README.md

jackycute:
To show "Community Edition" on title.


jackycute:
agreed, will add it first on index title

PR #690, opened 2018-01-20
Add option to enable `freely` permission in closed instance

### LINKED ISSUE #689: Share note to guests even if allowanonymous=false
Hi,

HackMD is a great tool, and I’m hosting it for myself and a few friends. I have allowanonymous=false and allowemailregister=false so that we can manually control the userbase. Setting allowanonymous=true makes guests able to *create* notes, which i want to forbid. However, having allowanonymous=false will remove the `freely` permission from the permission menu.

I think it is a legitimate usecase that only signed-in users can create a note, but guests cannot. When a signed-in user wants to share a note for a guest to edit, that should still be possible.

Thanks for your great tool nevertheless :+1: 
-Dario

Nebukadneza:
Some users may run a closed instance, having `allowanonymous=false` and no means of registration opened. For these instances, users may want to invite guests to notes that they [the users] created. It should be possible to allow such guests to either view or view+edit the notes that they received a link for.

This patch adds this functionality by adding a new config option, `allowanonymousedits`, which is only evaluated if `allowanonymous=false`. If a instance has `allowanonymous==false && allowanonymousedits==true`, then the `freely` permission will not be removed, allowing the behavior described above.

Fixes #689


daniele-athome:
I found this feature by chance, looking at issues/PR. [This is not documented here](https://hackmd.io/c/codimd-documentation/%2Fs%2Fcodimd-configuration#Privilege-and-Privacy). Am I looking in the wrong place? Thanks!

Issue #495, opened 2017-06-22
ASCII Math support


mlopezgva:
Hello!

***HackMD*** looks very promising, and I like to see so many _add-ons_ to the standard in the same way others have implemented them (kind of stantard) and other suggestions like ABC (I'm a musician, so I'm very excited with this one :smile:).

I would like present you [ASCIIMath](//www.asciimath.org),  a ligther way to write mathematical formulæ.
Please, take a look at my request [here](https://github.com/classeur/classeur/issues/195) for examples and comparison with LaTeX.

Thank you.

ccoenen:
Syntax looks nice!

PR #708, opened 2018-01-28
Support of AsciiMath

jackycute:
See #495

### DIFF
--- a/public/js/mathjax-config-extra.js
+++ b/public/js/mathjax-config-extra.js
@@ -4,5 +4,8 @@ window.MathJax = {
   tex2jax: {
     inlineMath: [['$', '$'], ['\\(', '\\)']],
     processEscapes: true
+  },
+  asciimath2jax: {
+    delimiters: [['$','$'], ['\\(','\\)']]
   }
 }
--- a/public/js/mathjax-config-extra.js
+++ b/public/js/mathjax-config-extra.js
@@ -6,6 +6,6 @@ window.MathJax = {
     processEscapes: true
   },
   asciimath2jax: {
-    delimiters: [['$','$'], ['\\(', '\\)']]
+    delimiters: [['$', '$'], ['\\(', '\\)']]
   }
 }
--- a/public/js/mathjax-config-extra.js
+++ b/public/js/mathjax-config-extra.js
@@ -6,6 +6,6 @@ window.MathJax = {
     processEscapes: true
   },
   asciimath2jax: {
-    delimiters: [['$','$'], ['\\(','\\)']]
+    delimiters: [['$','$'], ['\\(', '\\)']]
   }
 }
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "markdown-it-sub": "^1.0.0",
     "markdown-it-sup": "^1.0.0",
     "markdown-pdf": "^7.0.0",
-    "mathjax": "~2.7.0",
+    "mathjax": "~2.7.2",
     "mermaid": "~7.1.0",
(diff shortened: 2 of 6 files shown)


ccoenen:
I did not yet test this on any server, but as far as I can see this is limited to a 2.7.0 -> 2.7.2 upgrade. Does this indeed add AsciiMath support?

jackycute:
Hi @ccoenen 
I've tested on my local build.

Not just upgrade 2.7.0 to 2.7.2.
There are different includes of the config file `TeX-AMS-MML_HTMLorMML` to `TeX-MML-AM_HTMLorMML` which add support of AsciiMath.

Could you help to proof this too?

jackycute:
See more here: http://docs.mathjax.org/en/latest/config-files.html#the-tex-mml-am-htmlormml-configuration-file

SISheogorath:
I think there is some config missing:

https://github.com/hackmdio/hackmd/blob/aca5490b3a237dca88d8607806e89f86aad4c978/public/js/mathjax-config-extra.js#L1-L8

https://docs.mathjax.org/en/latest/options/preprocessors/asciimath2jax.html#configure-asciimath2jax

jackycute:
Thanks @SISheogorath 
But still can't be very sure about is this completely work or not.
Could anyone help?

public/js/mathjax-config-extra.js
```diff
@@ -4,5 +4,8 @@ window.MathJax = {
   tex2jax: {
     inlineMath: [['$', '$'], ['\\(', '\\)']],
     processEscapes: true
+  },
+  asciimath2jax: {
+    delimiters: [['$', '$'], ['\\(', '\\)']]
```
SISheogorath: I'm not sure they can be the same as tex2jax :/

I just checked the official default config and they are different there.
https://github.com/mathjax/MathJax/blob/master/config/default.js#L392-L398

By the way, my test is this:

```
sum_(i=1)^n i^3=((n(n+1))/2)^2
```

From http://asciimath.org/ which also recommends mathjax

jackycute:
Nah, I was wrong.
This can't support using the same syntax with TeX, they have different typesetting.
We could learn from draw.io for using different surround character to mixed them.
https://support.draw.io/questions/2949135/how-to-use-mathematical-typesetting

jackycute:
Since the <code>\`</code> actually overlap the code syntax of Markdown.
I would propose a new syntax that blend them in.
Use <code>\`</code> inside <code>$</code> would make Math using AsciiMath typesetting.
Like:
```
$`sum_(i=1)^n i^3=((n(n+1))/2)^2`$
```
Or
```
$$
`sum_(i=1)^n i^3=((n(n+1))/2)^2`
$$
```

Any ideas?

SISheogorath:
I'm not sure. maybe @ccoenen has an idea for this.

I could also imagine using something like `%%` because \` is definitely reserved from Markdown code blocks.

SISheogorath:
I think if we want to add ASCIImath support we should use `§§` which is uncommon enough to not causing issues and close enough to `$$`.

ccoenen:
Sorry, I have nothing of substance to contribute, I'm afraid.

jackycute:
We need new syntax or yaml metadata option to support AsciiMath.
It's ok to close this for now.

cben:
FWIW ``$`...`$`` is the delimiter for inline LaTeX math in GitLab markdown.

I'm not sure there is any syntax prior art for supporting both LaTeX and AsciiMath in markdown.

jackycute:
I don't think this will land in any close future since it's related to the syntax.
Close for now.

PR #716, opened 2018-02-01
don't require referer to find note id in socket.io connections (fixes #623)

### LINKED ISSUE #623: Getting the note ID from the Referer header is not reliable
The server seems to parse the `Referer` header to find the note ID. In my installation of Firefox Nightly and Chromium Dev Channel this does not work, as this header is not sent with the WebSocket connection. IMO it would be better if the note ID was explicitly sent when the WebSocket connection is brought up.

Relevant code is the extractNoteIdFromSocket function here:
https://github.com/hackmdio/hackmd/blob/ea9b1557f9834b2b140bd9b07488b5fca3953791/lib/realtime.js#L274-L285

stbuehler:
I've been running with this commit (based on d1d6d5810b12645ddb02275ce0c2498b2189a8a0) for quite some time and it didn't cause any problems so far (had some users with "paranoid" browser settings which removed the referer).

When this change gets deployed all users need to reload, as otherwise the javascript running in the browser doesn't send the `noteid` query parameter while the server is expecting it.

### DIFF
--- a/lib/realtime.js
+++ b/lib/realtime.js
@@ -272,16 +272,24 @@ function isReady () {
 }
 
 function extractNoteIdFromSocket (socket) {
-  if (!socket || !socket.handshake || !socket.handshake.headers) {
+  if (!socket || !socket.handshake) {
     return false
   }
-  var referer = socket.handshake.headers.referer
-  if (!referer) {
+  if (socket.handshake.query && socket.handshake.query.noteId) {
+    return socket.handshake.query.noteId
+  } else if (socket.handshake.headers) {
+    // this part is only for backward compatibility only; current code
+    // should be using noteId query parameter instead.
+    var referer = socket.handshake.headers.referer
+    if (!referer) {
+      return false
+    }
+    var hostUrl = url.parse(referer)
+    var noteId = config.urlpath ? hostUrl.pathname.slice(config.urlpath.length + 1, hostUrl.pathname.length).split('/')[1] : hostUrl.pathname.split('/')[1]
+    return noteId
+  } else {
     return false
   }
-  var hostUrl = url.parse(referer)
-  var noteId = config.urlpath ? hostUrl.pathname.slice(config.urlpath.length + 1, hostUrl.pathname.length).split('/')[1] : hostUrl.pathname.split('/')[1]
-  return noteId
 }
 
 function parseNoteIdFromSocket (socket, callback) {
--- a/public/js/index.js
+++ b/public/js/index.js
@@ -1762,6 +1762,9 @@ window.havePermission = havePermission
 var io = require('socket.io-client')
 var socket = io.connect({
   path: urlpath ? '/' + urlpath + '/socket.io/' : '',
+  query: {
+    noteId: noteid
+  },
   timeout: 5000, // 5 secs to timeout,
   reconnectionAttempts: 20 // retry 20 times on connect failed
 })


SISheogorath:
Thanks for your contribution! Looks pretty good, I checked it locally and besides the JS standard error, it works fine and is ready to get merged :)

Can you please fix the TravisCI issues: https://travis-ci.org/hackmdio/hackmd/jobs/336152849#L500-L501

lib/realtime.js
```diff
@@ -272,16 +272,10 @@ function isReady () {
...
     return false
   }
-  var referer = socket.handshake.headers.referer
-  if (!referer) {
-    return false
-  }
-  var hostUrl = url.parse(referer)
-  var noteId = config.urlpath ? hostUrl.pathname.slice(config.urlpath.length + 1, hostUrl.pathname.length).split('/')[1] : hostUrl.pathname.split('/')[1]
-  return noteId
+  return socket.handshake.query.noteid
```
ccoenen: should be `nodeId` (camelCase)

public/js/index.js
```diff
@@ -1762,6 +1762,9 @@ window.havePermission = havePermission
 var io = require('socket.io-client')
 var socket = io.connect({
   path: urlpath ? '/' + urlpath + '/socket.io/' : '',
+  query: {
+    noteid: noteid,
```
ccoenen: should be `nodeId` (camelCase) as well.

stbuehler:
I assume it was a typo and you meant `noteId`, not `nodeId`.  Trailing comma violating "standard".. oh well, I won't ever like JS anyway.  Removed the now unused `url` import.

ccoenen:
you're perfectly right. :-D Sorry for the typo in my typo correction.

jackycute:
Hi, guys!

This PR is cool.
But I have to mention that this is lacking of backward compatibility which I think it’s worth to keep.
Or you will need to announce breaking change and change the “minimalCompatibleVersion”.

ccoenen:
out of curiosity: are there clients that connect here other than HackMD itself?

stbuehler:
> But I have to mention that this is lacking of backward compatibility which I think it’s worth to keep.
> Or you will need to announce breaking change and change the “minimalCompatibleVersion”.

If necessary a fallback could be implemented.  Do you keep track of stuff you're going to drop with the next increment of `minimalCompatibleVersion` (because otherwise I don't think anyone is ever going to remember to clean this up)?

jackycute:
That “minimalCompatibleVersion” is designed for a real-time protocol changes like this. So it’s only needed to check when you modify something of the real-time scripts.

SISheogorath:
https://github.com/hackmdio/hackmd/blob/20242622004edc2438ecb6f59feab0b9eddc6280/lib/config/index.js#L21

I thought about the breaking change but didn't think of minimalCompatibleVersion as this look to me more like something for the Backend part. The client gets notified when the frontend version is changed and asks for refresh.

This change, as it changes the socket.io endpoint, breaks this update routine as it results in a 404. means people have to checkout their notes again. Not sure how to make this change compatible.

stbuehler:
So the version check is too late in this case, and we can't use it?

Well, keeping the old referer code as fallback should be easy - should I modify the PR that way?

SISheogorath:
Yes, I think we should keep some backwards compatibility.

lib/realtime.js
```diff
@@ -272,16 +272,24 @@ function isReady () {
 }
 
 function extractNoteIdFromSocket (socket) {
-  if (!socket || !socket.handshake || !socket.handshake.headers) {
+  if (!socket || !socket.handshake) {
     return false
   }
-  var referer = socket.handshake.headers.referer
-  if (!referer) {
+  if (socket.handshake.query) {
```
SISheogorath: We should evaluate the entire path, otherwise it'll may cause breaks.

`socket.handshake.query && socket.handshake.query.noteId`

ccoenen: this is implicitly done by the previous return (line 275 and 276). IF we end up here, we _know_ that sockt and socket handshake must exist.

ccoenen: On could argue that this breaks at some point down the road if this code gets modified by unsuspecting people. But that's what regression tests are for, IMO.

stbuehler: @SISheogorath You are right, it actually should break (due to query parameters like `transport`).

This time I actually tested the update, and the reconnect worked without reloading the page.

SISheogorath:
This feature really comes in time: https://blog.mozilla.org/security/2018/01/31/preventing-data-leaks-by-stripping-path-information-in-http-referrers/

We should release a new version before march.

PR #721, opened 2018-02-09
Fix field type to prevent data truncation of authorship

jackycute:
This PR solve similar data truncation issue for authorship field.
See #521 and #567 for details.


ccoenen:
What does `authorship` store? the regular Sequelize.TEXT is ~64KiB long, even in MySQL. Can the `authorship` field become that large?

Another remark: the migration should get a better filename (currently it's "unnamed").

jackycute:
@ccoenen Yes, authorship stores the JSON that describes about which part is edited by which author. So if there're many users collaborate in one note, this will definitely exceed 64KB.

I will fix that filename.

PR #732, opened 2018-02-19
Support ipsilon openidc

abbra:
This pull request adds support for signing into a Hackmd site with a custom identity provider based on [Ipsilon](https://ipsilon-project.org). This allows integrating Hackmd with identity management solutions like [FreeIPA](https://freeipa.org) or anything LDAP-based on POSIX systems via SSSD.

A benefit of doing so is in reusing existing identity management infrastructure that might be already deployed privately. Both Ipsilon and FreeIPA are free software.


abbra:
It looks like Travis is broken (the same test failed in an unrelated pull request as well).

SISheogorath:
Yes, unrelated to the pullrequest. Already saw it during the weekend. Will be fixed once I have time again. Then I should also find the time to review this (around next week).

I'm sorry that it takes so long.

abbra:
@SISheogorath  thanks. A second patch should probably get into a separate pull request, I'll do that.

SISheogorath:
Yes, can you please push it into an own branch? I really like your work!

abbra:
Done. Solved a conflict with README.md which was just pushed and moved anonymous editing into a separate pull request #734

abbra:
@SISheogorath could you please review this PR? It is sitting for three months already. I'm using this in production for my own setup all this time without any problems.

SISheogorath:
Sorry for the long delay. I already reviewed it earlier but didn't write stuff down.

We run into a problem right now, which is that there are more and more Id providers push into HackMD, invoke more dependencies and this way cause maintenance to become harder over time.

Can we change this PR to become a common OpenID provider. As Ipsilon uses OIDC anyways, we can also just build the generic way. This could simplify maintaining things a lot.

SISheogorath:
A more universal way to use OpenID was implemented by 9f9c4089be2112bfd8a55dd2cb996c811afab898

PR #734, opened 2018-02-19
Only show new guest note button i if anonymous editing is enabled

abbra:
Since there are two different settings: allowAnonymous and
allowAnonymousEdits, only show new guest note button in the latter case.

Same for publishing a note: if allowAnonymousEdits is not enabled, it makes no
sense to allow publishing a revision.

Signed-off-by: Alexander Bokovoy <ab@vda.li>

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -51,7 +51,9 @@ function responseError (res, code, detail, msg) {
     code: code,
     detail: detail,
     msg: msg,
-    useCDN: config.usecdn
+    useCDN: config.usecdn,
+    allowAnonymousEdits: false,
+    signin: false
   })
 }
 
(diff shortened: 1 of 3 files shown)


lib/response.js
```diff
@@ -51,7 +51,9 @@ function responseError (res, code, detail, msg) {
     code: code,
     detail: detail,
     msg: msg,
-    useCDN: config.usecdn
+    useCDN: config.usecdn,
+    allowAnonymousEdits: false,
+    signin: false
```
SISheogorath: Shouldn't these newly added options get a useful value instead of being statically set to false?

abbra: in their context we only have a response object, `res`. If that object can allow them to be derived from something -- fine. I was not able to find out what I can derive them from.

As this is just about displaying an error message, I think it is OK to not display the edit/publish buttons but we have to have variables defined to avoid throwing an error in the EJS code that references them.

PR #740, opened 2018-02-25
Extend HTML5 support by whitelisting various tags

### LINKED ISSUE #671: [Question] Any plans of supporting more HTML tags?
The most interesting feature in HackMD that intrigues me, is the partial support of HTML. You can insert a line of HTML in the document directly to implement some features that markdown does not provide.

So the question is, how many HTML tags can be used in HackMD directly? And for the tags that are not supported, are there any plans of supporting them?

For example, I can easily insert an audio/video/iframe element, but SVG/Canvas elements will be treated as text. And then I try the ruby tag:

```html
<ruby>兎<rt>うさぎ</rt></ruby>
```

Well it can be handled in Github Markup: <ruby>兎<rt>うさぎ</rt></ruby>, but in HackMD it will be treated as text, too. So this issue is also a feature request issue. 😂

SISheogorath:
HTML5 provides a wide feature set of useful elements. Since Markdown
usually supports HTML it should be able to use these HTML5 tags as well.
As they were requested by some users and they where checked for being
safe, whitelisting them isn't a problem. To make the experience the same
as on GitHub when it comes to the basic look and feel of the rendered
markdown, some CSS was added to make the summary and the details tag
look like on GitHub.

---

Fixes #671

### DIFF
--- a/public/css/markdown.css
+++ b/public/css/markdown.css
@@ -188,3 +188,12 @@
 .markdown-body .alert > ul {
     margin-bottom: 0;
 }
+
+/* Make details boxes look like on GitHub */
+.markdown-body details summary {
+    cursor: pointer;
+}
+
+.markdown-body summary {
+    display: list-item;
+}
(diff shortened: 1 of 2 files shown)


jackycute:
The details tag already allow in the default settings
https://github.com/leizongmin/js-xss/blob/master/lib/default.js#L34

As you specified here, it lacks the attribute `open`.
I would suggest to take it out.

PR #750, opened 2018-03-01
Use ldap.usernameField over hardcoded uid fields

fooker:
In our environment we don't have `uid` set on users. This leads to all users share the same profile called `LDAP-undefined`.

This fixes this under the assumption that usernames are unique anyways.

### DIFF
--- a/lib/web/auth/ldap/index.js
+++ b/lib/web/auth/ldap/index.js
@@ -24,8 +24,11 @@ passport.use(new LDAPStrategy({
   }
 }, function (user, done) {
   var uuid = user.uidNumber || user.uid || user.sAMAccountName
-  var username = uuid
+  if (config.ldap.useridField && user[config.ldap.useridField]) {
+    uuid = user[config.ldap.useridField]
+  }
 
+  var username = uuid
   if (config.ldap.usernameField && user[config.ldap.usernameField]) {
     username = user[config.ldap.usernameField]
   }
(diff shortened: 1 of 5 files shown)


lib/web/auth/ldap/index.js
```diff
@@ -23,15 +23,13 @@ passport.use(new LDAPStrategy({
     tlsOptions: config.ldap.tlsOptions || null
   }
 }, function (user, done) {
-  var uuid = user.uidNumber || user.uid || user.sAMAccountName
-  var username = uuid
-
+  var uusername = user.uidNumber || user.uid || user.sAMAccountName
```
SISheogorath: `uusername`? probably a typo.

fooker:
Good point. Should the setting fall back to the default of `user.uidNumber || user.uid || user.sAMAccountName` or should it be a required field like username is?

fooker:
Ok, changed it to use a new config option which defaults to `uid`.
I'm still not sure how to handle the misconfiguration case as all users without an field will still share the same profile.

lib/web/auth/ldap/index.js
```diff
@@ -23,15 +23,15 @@ passport.use(new LDAPStrategy({
     tlsOptions: config.ldap.tlsOptions || null
   }
 }, function (user, done) {
-  var uuid = user.uidNumber || user.uid || user.sAMAccountName
-  var username = uuid
+  var uid = user[config.ldap.useridField || 'uid']
```
SISheogorath: I think we still should fall back to the old defaults because we don't expect our users to change this on update. So old settings shouldn't change anything.

fooker:
Here you go.

PR #756, opened 2018-03-06
Remove engine.io-client dependency

### LINKED ISSUE #715: engine.io-client dependency broken, socket.io connections not working
In a fresh install:

Right now the `socket.io-client@^2.0.4` dependency depends on `engine.io-client@~3.1.0`, ending up as `3.1.4`, and the `socket.io@~2.0.4` pulls in `engine.io@~3.1.0` (again ending up as `3.1.4`).

But `hackmd` depends on `engine.io-client@^1.8.2`, ending up as `1.8.5` - which does not work with `engine.io@3.1.4`.

I think the `engine.io-client@^1.8.2` dependency should be removed, and `modulesDirectories` should be set to only contain `node_modules`, not absolute paths (see https://github.com/webpack/webpack/issues/472#issuecomment-55609294); that way webpack should still find `engine.io-client`.

It seems webpack needs more memory and more time that way, but maybe you can figure out why (perhaps its including some modules more than once, and the dependencies should get fixed further to have only one version of those).

As a start increasing the `engine.io-client` dependency to `engine.io-client@^3.1.0` should workaround the problem, but that is NOT a longterm solution!

davidmehren:
This is my attempt to fix #715.
After removing the `engine.io-client@^1.8.2` dependency as @stbuehler suggested, `raphael` needed some convincing to still find `eve`.

In my (admittedly limited) testing, HackMD now works well, even with `"usecdn": false`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,6 @@
     "diff-match-patch": "git+https://github.com/hackmdio/diff-match-patch.git",
     "ejs": "^2.5.5",
     "emojify.js": "~1.1.0",
-    "engine.io-client": "^1.8.2",
     "express": ">=4.14",
     "express-session": "^1.14.2",
     "file-saver": "^1.3.3",
--- a/webpackBaseConfig.js
+++ b/webpackBaseConfig.js
@@ -355,10 +355,7 @@ module.exports = {
   },
 
   resolve: {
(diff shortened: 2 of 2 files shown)


SISheogorath:
Thanks a lot for your work!

PR #758, opened 2018-03-07
Change config to camel case with backwards compatibility

### LINKED ISSUE #315: Inconsistent cases in config
In my understanding the configuration keys (in `config.json`) are case sensitive. Unfortunately, they don't use case consistently. `allowanonymous` is all lowercase while `imageUploadType` is camel case.

I'd suggest to convert everything to camel case due its readability. Backward compatibility could be achieved by e.g. changing
```js
var allowanonymous = process.env.HMD_ALLOW_ANONYMOUS 
    ? (process.env.HMD_ALLOW_ANONYMOUS === 'true') 
    : ((typeof config.allowanonymous === 'boolean') ? config.allowanonymous : true);
```
to 
```js
var allowanonymous = process.env.HMD_ALLOW_ANONYMOUS 
    ? (process.env.HMD_ALLOW_ANONYMOUS === 'true') 
    : (((typeof config.allowAnonymous === 'boolean') ? config.allowAnonymous : true) 
        || ((typeof config.allowanonymous === 'boolean') ? config.allowanonymous : true));
```
Et cetera.

Definitely low priority but would make configuration / initial setup more intuitive.

SISheogorath:
This refactors the configs a bit to now use camelcase everywhere.
This change should help to clean up the config interface and make it
better understandable.

fixes #315

### DIFF
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -27,6 +27,7 @@ const fileConfig = fs.existsSync(configFilePath) ? require(configFilePath)[env] 
 
 let config = require('./default')
 merge(config, require('./defaultSSL'))
+merge(config, require('./oldDefault'))
 merge(config, debugConfig)
 merge(config, packageConfig)
 merge(config, fileConfig)
@@ -51,35 +52,35 @@ if (config.ldap.tlsca) {
 
 // Permission
 config.permission = Permission
-if (!config.allowanonymous && !config.allowanonymousedits) {
+if (!config.allowAnonymous && !config.allowAnonymousedits) {
   delete config.permission.freely
 }
-if (!(config.defaultpermission in config.permission)) {
-  config.defaultpermission = config.permission.editable
+if (!(config.defaultPermission in config.permission)) {
+  config.defaultPermission = config.permission.editable
 }
 
 // cache result, cannot change config in runtime!!!
 config.isStandardHTTPsPort = (function isStandardHTTPsPort () {
-  return config.usessl && config.port === 443
+  return config.useSSL && config.port === 443
 })()
 config.isStandardHTTPPort = (function isStandardHTTPPort () {
-  return !config.usessl && config.port === 80
+  return !config.useSSL && config.port === 80
 })()
 
 // cache serverURL
-config.serverurl = (function getserverurl () {
+config.serverURL = (function getserverurl () {
   var url = ''
   if (config.domain) {
-    var protocol = config.protocolusessl ? 'https://' : 'http://'
+    var protocol = config.protocolUseSSL ? 'https://' : 'http://'
     url = protocol + config.domain
-    if (config.urladdport) {
+    if (config.urlAddPort) {
       if (!config.isStandardHTTPPort || !config.isStandardHTTPsPort) {
         url += ':' + config.port
       }
     }
   }
-  if (config.urlpath) {
-    url += '/' + config.urlpath
+  if (config.urlPath) {
+    url += '/' + config.urlPath
   }
   return url
 })()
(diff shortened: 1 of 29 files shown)


SISheogorath:
@ccoenen any thoughts on this?

lib/config/index.js
```diff
@@ -96,11 +97,21 @@ config.isGitLabEnable = config.gitlab.clientID && config.gitlab.clientSecret
...
+for (let i = keys.length; i--;) {
+  const uppercase = /[A-Z]/
+  // if the config contains uppercase letters
+  // and a lowercase version of this setting exists
+  // and the original, new, config with uppercase is not set
+  // in this case we set the new config using the old key.
+  if (uppercase.test(keys[i]) &&
+    config[keys[i].toLowerCase()] &&
+    !config[keys[1]]) {
+    config[keys[i]] = config[keys[i].toLowerCase()]
```
ccoenen: This should also print a warning.

    `Warning: config.js contains deprecated lowercase setting for ${keys[i]}. Please change your config.js file to replace ${keys[i].toLowerCase()} with ${keys[i]}`

This way, we could remove the section at some point in the future. (think: warn for a year, break with warning for a year, remove completely)
Another possibility would be, to use this part to compare config.js' keys with a set of default keys - I see one upside and one downside to this:
- You now catch misspelling in the config
- You can't simply add a new config without also remembering that there's a default config file or config keys set.

lib/config/index.js
```diff
@@ -96,11 +97,21 @@ config.isGitLabEnable = config.gitlab.clientID && config.gitlab.clientSecret
...
 config.isSAMLEnable = config.saml.idpSsoUrl
-config.isPDFExportEnable = config.allowpdfexport
+config.isPDFExportEnable = config.allowPDFExport
 
 // merge legacy values
-if (config.imageUploadType && !config.imageuploadtype) {
-  config.imageuploadtype = config.imageUploadType
+let keys = Object.keys(config)
+for (let i = keys.length; i--;) {
+  const uppercase = /[A-Z]/
```
ccoenen: `const uppercase....` can be taken out of the loop.

lib/config/oldDefault.js
```diff
@@ -0,0 +1,42 @@
+'use strict'
```
ccoenen: do we need this file any longer?

lib/config/index.js
```diff
@@ -96,11 +97,21 @@ config.isGitLabEnable = config.gitlab.clientID && config.gitlab.clientSecret
...
+  // if the config contains uppercase letters
+  // and a lowercase version of this setting exists
+  // and the original, new, config with uppercase is not set
+  // in this case we set the new config using the old key.
+  if (uppercase.test(keys[i]) &&
+    config[keys[i].toLowerCase()] &&
```
ccoenen: Through the following addition, we'll be using `keys[i].toLowerCase()` quite a bit, so it's probably a good idea to `let lowercaseCandidate = keys[i].toLowerCase()` this.

lib/response.js
```diff
@@ -45,22 +45,22 @@ var response = {
...
-  res.render(config.indexpath, {
-    url: config.serverurl,
-    useCDN: config.usecdn,
-    allowAnonymous: config.allowanonymous,
-    allowAnonymousEdits: config.allowanonymousedits,
+  res.render(config.indexPath, {
+    url: config.serverURL,
+    useCDN: config.useCDN,
+    allowAnonymous: config.allowAnonymous,
+    allowAnonymousEdits: config.allowAnonymousedits,
```
ccoenen: these are now very nice!

lib/response.js
```diff
@@ -72,8 +72,8 @@ function showIndex (req, res, next) {
     ldapProviderName: config.ldap.providerName,
     saml: config.isSAMLEnable,
     email: config.isEmailEnable,
-    allowemailregister: config.allowemailregister,
-    allowpdfexport: config.allowpdfexport,
+    allowemailregister: config.allowEmailRegister,
+    allowpdfexport: config.allowPDFExport,
```
ccoenen: can these *keys* be changed without breaking anything?

lib/response.js
```diff
@@ -107,24 +107,24 @@ function responseHackMD (res, note) {
     ldapProviderName: config.ldap.providerName,
     saml: config.isSAMLEnable,
     email: config.isEmailEnable,
-    allowemailregister: config.allowemailregister,
-    allowpdfexport: config.allowpdfexport
+    allowemailregister: config.allowEmailRegister,
+    allowpdfexport: config.allowPDFExport
```
ccoenen: same as above: the keys are all lowercase, not sure if camelCasing would break anything.

ccoenen:
I did not yet run this in production, I also did not go on a hunt for "forgotton" options in the code or something. Right now, this is just a review of the changeset.

lib/config/index.js
SISheogorath: :+1:

lib/config/index.js
SISheogorath: :+1:

lib/response.js
SISheogorath: For now I stayed on the server side :D But sure we can extend this to the client side as well ^^

lib/config/oldDefault.js
SISheogorath: We probably don't need it. I kept it to either migrate and also keep it consistent.

lib/config/index.js
SISheogorath: Yes, a warning should be added. :+1: 

I'll introduce a way to ignore invalid values and jump back to useful defaults in a future commit outside of this PR. for now I would like to transit to the camel case way of writing :)

SISheogorath:
- Fixed missing options in environment.js
- Added requested changes
- Fixed template

SISheogorath:
@ccoenen Have another look at this?

lib/config/index.js
```diff
@@ -97,21 +98,33 @@ config.isGitLabEnable = config.gitlab.clientID && config.gitlab.clientSecret
...
 // merge legacy values
-if (config.imageUploadType && !config.imageuploadtype) {
-  config.imageuploadtype = config.imageUploadType
+let keys = Object.keys(config)
+const uppercase = /[A-Z]/
+for (let i = keys.length; i--;) {
+  let lowercaseKey = keys[i].toLowerCase()
+  // if the config contains uppercase letters
+  // and a lowercase version of this setting exists
+  // and the original, new, config with uppercase is not set
```
ccoenen: this explanation is hard to read. I would expect this to read:
`and the correct config with uppercase is not set`

Reason: the "new" will not make sense to anyone but us. It's also weird that original and new refer to the same thing.

PR #761, opened 2018-03-10
Add config option for report URI in CSP

SISheogorath:
This option is needed as it's currently not possible to add an report
URI by the directives array. This option also allows to get CSP reports
not only on docker based setup but also on our heroku instances.

Improves CSP support #594

### DIFF
--- a/lib/csp.js
+++ b/lib/csp.js
@@ -30,6 +30,7 @@ CspStrategy.computeDirectives = function () {
     addInlineScriptExceptions(directives)
   }
   addUpgradeUnsafeRequestsOptionTo(directives)
+  addReportURI(directives)
   return directives
 }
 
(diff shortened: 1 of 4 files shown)


lib/csp.js
```diff
@@ -72,6 +73,12 @@ function addUpgradeUnsafeRequestsOptionTo (directives) {
   }
 }
 
+function addReportURI (directives) {
+  if (config.csp.reportURI) {
+    directives.reportURI = config.csp.reportURI
```
xxyy: Does this work? The [docs](https://helmetjs.github.io/docs/csp/) have `reportUri` and fields are case-sensitive, right?

I think it would make sense to use `reportUri` for the config option as well, for consistency with their format – that way we can't have confusions with their documentation.

SISheogorath: Agreed. But yes, it worked ^^

Issue #767, opened 2018-03-15
parseInt(HMD_MINIO_PORT) cause another bug


vazontang:
####  Sorry for that my PR #765 does't fix bug #763 perfectly.

The code I wrote in `/lib/config/environment.js`  :
```javascript
port: parseInt(process.env.HMD_MINIO_PORT)
```
Cause  another bug which make  `config.minio.port`   neither ***default value*** nor ***value in config.json*** while **HMD_MINIO_PORT** hasn't been defined  .

The reason is ***environment.js***  is merged after than ***config.json***  in  `/lib/config/index.js` .

In addition ,  `lodash.merge` [only skip ***undefined*** value](https://lodash.com/docs#merge) .

However using `port: parseInt(process.env.HMD_MINIO_PORT)` make value of port ***isn't  undefined  but  NaN***  while```process.env.HMD_MINIO_PORT```  is **undefined** .

And It makes  `config.minio.port` overwrite with ***NaN***  even though **HMD_MINIO_PORT** hasn't been set in env  . 

---

### Value of port should be **undefined** when **HMD_MINIO_PORT** isn't set in ENV to prevent original  value be overwrote  .   And I rewrite code into this to fix it :
```javascript
port: (process.env.HMD_MINIO_PORT == undefined) ? undefined : parseInt(process.env.HMD_MINIO_PORT)
```

PR #768, opened 2018-03-16
Fix minio port been NaN while  HMD_MINIO_PORT  isn't set in ENV #767

vazontang:
fix hackmdio/hackmd#767
---
`port: parseInt(process.env.HMD_MINIO_PORT)` will make `config.minio.port` be ***NaN*** while `HMD_MINIO_PORT` hasn't set in ENV .   And It brings on that port value in `config.json` doesn't work because `lodash.merge` only skip  **undefined** value .

Therefore , value of minio port should be **undefined** when `HMD_MINIO_PORT` isn't set to prevent original value be overwrote with **NaN** .


SISheogorath:
I'm sorry for jumping in here, but as it actually annoyed me how the last PR broke stuff, I decided to fix it myself in a clean way. See #769

PR #770, opened 2018-03-17
Add check for undefined UUID

### LINKED ISSUE #764: Misconfigured LDAP authenticator leaks private documents
When the LDAP authenticator is misconfigured to use a non-existing field to determine the user ID, all users share the same account and private documents are leaked to other users.

The LDAP authenticator (and other authenticators) should never authenticate a user if it is misconfigured.

SISheogorath:
This check is needed at there are tons of LDAP implementations out there
and none has at least one guaranteed unique field. As we currently check
three fields and added an option to select one yourself, it's still not
said that any of these fields is set. This will now create an error
and fail the authentication instead of letting people may get access to
other people's notes which are stored under a this way deterministic
wrong userid named `LDAP-undefined`.

Fixes #764 

@fooker would you like to review and test this?

### DIFF
--- a/lib/web/auth/ldap/index.js
+++ b/lib/web/auth/ldap/index.js
@@ -23,9 +23,16 @@ passport.use(new LDAPStrategy({
     tlsOptions: config.ldap.tlsOptions || null
   }
 }, function (user, done) {
-  var uuid = user.uidNumber || user.uid || user.sAMAccountName
+  var uuid = user.uidNumber || user.uid || user.sAMAccountName || undefined
   if (config.ldap.useridField && user[config.ldap.useridField]) {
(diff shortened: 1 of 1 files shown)


lib/web/auth/ldap/index.js
```diff
@@ -33,6 +33,13 @@ passport.use(new LDAPStrategy({
     username = user[config.ldap.usernameField]
   }
 
+  if (typeof uuid === 'undefined') {
```
fooker: Maybe this should be moved up before the username code. If the username field is not used, this will say something like `Could not determine UUID for LDAP user "undefined"`.

fooker:
Looks got to me.

PR #771, opened 2018-03-18
Refactoring imageRouter to modularity

SISheogorath:
This should make the imageRouter more modular and easier to extent. Also
a lot of code duplication was removed which should simplify maintenance
in future.

In the new setup we only need to provide a new module file which exports
a function called `uploadImage` and takes a filePath and a callback as
argument. The callback itself takes an error and an url as parameter.
This eliminates the need of a try-catch-block around the statement and
re-enabled the optimization in NodeJS.


ccoenen:
nice!

PR #782, opened 2018-03-26
Add title attribute in table of contents

### LINKED ISSUE #781: Feature request: Popup Outline Title
screen shot 2018-03-26 at 9 01 43 am

The right-side-outline help hackmders quick jump to the topic, I really love it!
But the long title will be cut(...). Sometimes we can't get what it means

Is it possible to pop up the full title? (maybe use "title" props)

Thanks for your effort!

SISheogorath:
Right now the full title of an element is may not shown as the space of
the ToC is limited. With this path it'll be shower on hover and this way
provide more useful information.

Fixes #781

### DIFF
--- a/public/vendor/md-toc.js
+++ b/public/vendor/md-toc.js
@@ -54,6 +54,7 @@
       var j = i + 1
       this._elTitleElement = this.elTitleElements[i]
       this._elTitleElementName = this._elTitleElement.tagName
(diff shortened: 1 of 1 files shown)


SISheogorath:
It's probably not the best thing to write a patch in the morning before you are completely awake ^^

Thanks @ccoenen for pointing me to this huge attack vector. Fixed now! :+1:

PR #783, opened 2018-03-26
Split authentication guides into multiple documents

SISheogorath:
Splitting the documentation should provide an easier access to the
documentation people searching for and result in less merge conflicts
when adding new documentation here.


ccoenen:
do we have documentation on local accounts? If not: this should at least be mentioned somewhere close to these methods.

SISheogorath:
Right now we don't have any docs about local accounts.

Anyone wants to write them? :D

PR #784, opened 2018-03-26
Add "generic" OAuth2 support

pferreir:
The currently available OAuth methods are limited to a few providers. This PR implements a "generic" OAuth workflow which will fetch user profile info from a given JSON endpoint and extract a set of configurable fields. Example:

```json
        "oauth2": {
          "clientID": "test_hackmd_notasbad",
          "clientSecret": "lc05GbmUT5crF33341CcpDbliG8KNW0HS4p9sS91Vx81",
          "authorizationURL": "https://authprovider/oauth/authorize",
          "tokenURL": "https://authprovider/oauth/token",
          "userProfileURL": "https://authprovider/api/user",
          "userProfileUsernameAttr": "username",
          "userProfileDisplayNameAttr": "data.displayName",
          "userProfileEmailAttr": "email"
        }
```

I tried to guide myself by the existing OAuth provider definitions, please let me know if there are any better ways of doing what I'm trying to do.

### DIFF
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -98,6 +98,7 @@ config.isGitLabEnable = config.gitlab.clientID && config.gitlab.clientSecret
 config.isMattermostEnable = config.mattermost.clientID && config.mattermost.clientSecret
 config.isLDAPEnable = config.ldap.url
 config.isSAMLEnable = config.saml.idpSsoUrl
+config.isOAuth2Enable = config.oauth2.clientID && config.oauth2.clientSecret
 config.isPDFExportEnable = config.allowPDFExport
 
 // merge legacy values
--- a/lib/models/user.js
+++ b/lib/models/user.js
@@ -26,10 +26,10 @@ module.exports = function (sequelize, DataTypes) {
       type: DataTypes.TEXT
     },
     accessToken: {
-      type: DataTypes.STRING
+      type: DataTypes.TEXT
     },
     refreshToken: {
-      type: DataTypes.STRING
+      type: DataTypes.TEXT
     },
     email: {
       type: Sequelize.TEXT,
--- a/public/views/shared/signin-modal.ejs
+++ b/public/views/shared/signin-modal.ejs
@@ -48,7 +48,12 @@
                     <i class="fa fa-users"></i> <%= __('Sign in via %s', 'SAML') %>
                 </a>
                 <% } %>
-                <% if((facebook || twitter || github || gitlab || mattermost || dropbox || google || saml) && ldap) { %>
+                <% if(oauth2) { %>
+                <a href="<%- url %>/auth/oauth2" class="btn btn-lg btn-block btn-social btn-soundcloud">
+                    <i class="fa fa-mail-forward"></i> <%= __('Sign in via %s', oauth2ProviderName || 'OAuth2') %>
+                </a>
+                <% } %>
+                <% if((facebook || twitter || github || gitlab || mattermost || dropbox || google || saml || oauth2) && ldap) { %>
                 <hr>
                 <% }%>
                 <% if(ldap) { %>
@@ -73,7 +78,7 @@
                     </div>
                 </form>
                 <% } %>
-                <% if((facebook || twitter || github || gitlab || mattermost || dropbox || google || ldap) && email) { %>
+                <% if((facebook || twitter || github || gitlab || mattermost || dropbox || google || ldap || oauth2) && email) { %>
                 <hr>
                 <% }%>
                 <% if(email) { %>
--- a/lib/config/default.js
+++ b/lib/config/default.js
@@ -72,6 +72,12 @@ module.exports = {
   },
   s3bucket: undefined,
   // authentication
+  oauth2: {
+    authorizationURL: undefined,
+    tokenURL: undefined,
+    clientID: undefined,
+    clientSecret: undefined
+  },
   facebook: {
     clientID: undefined,
     clientSecret: undefined
--- a/lib/config/environment.js
+++ b/lib/config/environment.js
@@ -66,6 +66,17 @@ module.exports = {
     clientID: process.env.HMD_MATTERMOST_CLIENTID,
     clientSecret: process.env.HMD_MATTERMOST_CLIENTSECRET
   },
+  oauth2: {
+    baseURL: process.env.HMD_OAUTH2_BASEURL,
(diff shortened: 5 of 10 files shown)


public/views/shared/signin-modal.ejs
```diff
@@ -48,7 +48,12 @@
                     <i class="fa fa-users"></i> <%= __('Sign in via %s', 'SAML') %>
                 </a>
                 <% } %>
-                <% if((facebook || twitter || github || gitlab || mattermost || dropbox || google || saml) && ldap) { %>
+                <% if(oauth2) { %>
+                <a href="<%- url %>/auth/oauth2" class="btn btn-lg btn-block btn-social btn-soundcloud">
+                    <i class="fa fa-mail-forward"></i> <%= __('Sign in via %s', 'CERN SSO') %>
```
SISheogorath: The name should be configurable.

lib/models/user.js
```diff
@@ -26,10 +26,10 @@ module.exports = function (sequelize, DataTypes) {
       type: DataTypes.TEXT
     },
     accessToken: {
-      type: DataTypes.STRING
+      type: DataTypes.TEXT
```
SISheogorath: Any reason why this change is needed?

lib/config/index.js
```diff
@@ -99,6 +99,7 @@ config.isMattermostEnable = config.mattermost.clientID && config.mattermost.clie
 config.isLDAPEnable = config.ldap.url
 config.isSAMLEnable = config.saml.idpSsoUrl
 config.isPDFExportEnable = config.allowPDFExport
+config.isOAuth2Enable = config.oauth2.clientID && config.oauth2.clientSecret
```
SISheogorath: Can we move this up to the other auth providers?

lib/models/user.js
pferreir: Sorry, I should have mentioned this. Our OAuth provider sometimes produces tokens that are longer that 255 chars (which I believe is the maximum STRING length). From [what I understood](https://stackoverflow.com/a/16365828/682095) there isn't a maximum lenght we can assume will be respected.

SISheogorath:
@pferreir do you run it in your organisation?

If so, could you please build this branch using the docker-hackmd setup and verify that it's working? Then I'll merge it :)

Kemyke:
I can successfully build a version from @pferreir branch.
Here is my docker-compose.yml:
```
# Using version 3 to provide play-with-docker badge
# You can change to version 2 without breaking.
#version: '2'
version: '3'
services:
  database:
    # Don't upgrade PostgreSQL by simply changing the version number
    # You need to migrate the Database to the new PostgreSQL version
    image: postgres:9.6-alpine
    environment:
      - POSTGRES_USER=hackmd
      - POSTGRES_PASSWORD=hackmdpass
      - POSTGRES_DB=hackmd
    volumes:
      - database:/var/lib/postgresql/data
    networks:
      backend:
    restart: always

  # MySQL example
  # Most of the documentation that applies to PostgreSQL applies also to MySQL
  #database:
  #    # You should be able to upgrade MySQL without problems
  #    # but to make sure no even when a problem appears you
  #    # should have a backup
  #    image: mariadb:10
  #    environment:
  #      - MYSQL_USER=hackmd
  #      - MYSQL_PASSWORD=hackmdpass
  #      - MYSQL_DATABASE=hackmd
  #      - MYSQL_ALLOW_EMPTY_PASSWORD=true
  #    volumes:
  #      - database:/var/lib/mysql
  #      # This config provides UTF-8 support to the database by default
  #      # If this config is not used, HackMD breaks as it tries to write
  #      # UTF-8 to a latin database.
  #      - ./resources/utf8.cnf:/etc/mysql/conf.d/utf8.cnf
  #    networks:
  #      backend:
  #    restart: always

  app:
    # Uncomment the following section to build the image yourself:
    build:
      context: .
      dockerfile: debian/Dockerfile
      args:
        - "VERSION=add-oauth2-support"
        - "HACKMD_REPOSITORY=https://github.com/pferreir/hackmd.git"
    #image: hackmdio/hackmd:1.1.0-ce
    environment:
      # DB_URL is formatted like: <databasetype>://<username>:<password>@<hostname>/<database>
      # Other examples are:
      # - mysql://hackmd:hackmdpassdatabase:3306/database
      # - sqlite:///data/sqlite.db (NOT RECOMMENDED)
      # - For details see the official sequelize docs: http://docs.sequelizejs.com/en/v3/
      - HMD_DB_URL=postgres://hackmd:hackmdpass@database:5432/hackmd
      - HMD_OAUTH2_TOKEN_URL=https://login.microsoftonline.com/<tenantId>/oauth2/token
      - HMD_OAUTH2_AUTHORIZATION_URL=https://login.microsoftonline.com/<tenantId>/oauth2/authorize?resource=https%3A%2F%2Fgraph.windows.net
      - HMD_OAUTH2_CLIENT_ID=<AppId>
      - HMD_OAUTH2_CLIENT_SECRET=<AppSecret>
      - HMD_OAUTH2_USER_PROFILE_URL=https://graph.windows.net/me?api-version=1.6
      - HMD_OAUTH2_USER_PROFILE_USERNAME_ATTR=mailNickname
      - HMD_OAUTH2_USER_PROFILE_DISPLAY_NAME_ATTR=displayName
      - HMD_OAUTH2_USER_PROFILE_EMAIL_ATTR=mail
    ports:
      # Ports that are published to the outside.
      # The latter port is the port inside the container. It should always stay on 3000
      # If you only specify a port it'll published on all interfaces. If you want to use a
      # local reverse proxy, you may want to listen on 127.0.0.1.
      # Example:
      # - "127.0.0.1:3000:3000"
      - "3000:3000"
    networks:
      backend:
    restart: always
    depends_on:
      - database

# Define networks to allow best isolation
networks:
  # Internal network for communication with PostgreSQL/MySQL
  backend:

# Define named volumes so data stays in place
volumes:
  # Volume for PostgreSQL/MySQL database
  database:
```
With this setting i can create an environment and login to our AzureAD through OAuth2. 



It is not really clear why a separate API call is required to get the profile data. All of that data is available in the token. So maybe the configuration can be simplified in the future. For example Grafane use the data from token if no profile endpoint is given, it is a very convinient default IMHO.

pferreir:
@SISheogorath That is actually the code we are using right now at my organization. It works here :)

SISheogorath:
Thanks a lot for your work!

This should help to keep dependencies minimal and this way make HackMD safer!

PR #790, opened 2018-03-29
Fix modal and panel colors in night mode

SISheogorath:
Night mode provides a generally, dark interface. This fix provides the
needed CSS to also turn modal and panels into night mode design as well.
This mainly effects the help modal.


SISheogorath:
Added fixes for code blocks

PR #791, opened 2018-03-30
Fix CSP for disqus and Google Analytics

SISheogorath:
This commit should fix existing problems with Disqus and Google
Analytics enabled in the meta-yaml section of a note.

Before this commit they were blocked by the strict CSP. It's still
possible to disable the added directives using `addDisqus` and
`addGoogleAnalytics` in the `csp` config section.

They are enabled by default to prevent breaking changes.

--- 

**GA policy is still untested**

/cc @xxyy

### DIFF
--- a/config.json.example
+++ b/config.json.example
@@ -27,7 +27,9 @@
             "directives": {
             },
             "upgradeInsecureRequests": "auto"
-            "addDefaults": true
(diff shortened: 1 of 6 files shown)


SISheogorath:
I'll merge it for now, if it's still broken, feel free to complain.

PR #797, opened 2018-04-09
Add check for noteId length

SISheogorath:
As we know the length of an UUID we can check if the base64 string 
of the provided UUID is long enough for a legacy base64 encoded
nodeId and stop processing it in legacy mode, if it's not the case.

This should make the ugly warning way less common.

/cc @dekonnection

### DIFF
--- a/lib/models/note.js
+++ b/lib/models/note.js
@@ -211,6 +211,15 @@ module.exports = function (sequelize, DataTypes) {
           },
           // parse note id by LZString is deprecated, here for compability
           parseNoteIdByLZString: function (_callback) {
+            // Calculate minimal string length for an UUID that is encoded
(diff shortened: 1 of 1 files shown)


lib/models/note.js
```diff
@@ -211,6 +211,15 @@ module.exports = function (sequelize, DataTypes) {
           },
           // parse note id by LZString is deprecated, here for compability
           parseNoteIdByLZString: function (_callback) {
+            // Calculate mainimal string length for an UUID that is encoded
```
jackycute: typo here I think?
`mainimal` -> `minimal`

SISheogorath: Fixed. Thanks :)

PR #830, opened 2018-05-25
GDPR compliant part 1

SISheogorath:
As GDPR kicked in today, we are pretty late on this.

It should provide some basic features that allow to run HackMD fine for GDPR. Not more, not less. Just enough for now.

When the next big re-factoring for user management comes, this should be redone.


jackycute:
@SISheogorath Here you might miss a sign-off in the last commit.

PR #833, opened 2018-06-01
Fix callback validation

ahoka:
Don't call the callback when it's undefined or not a function but log the error.


SISheogorath:
As soon as it's signed off I will merge this :)

`git commit --amend --signoff` ;)

ahoka:
Sorry I forgot. Here it is.

PR #849, opened 2018-06-22
turn concatenated string into a multi-line template string

jekrb:
👋 

I like hackmd a lot and was looking to be helpful with development!
I noticed that the JS seemed to follow ES6 conventions such as `import` and arrow functions. So for a start, I thought I would change this turn this block of code into its ES6 alternative (from concat string into multi-line string with backtics).  I'm looking into other ways to contribute as well, and this was a quick start for me, no hard feelings if this PR doesn't get merged xD

### DIFF
--- a/public/js/cover.js
+++ b/public/js/cover.js
@@ -34,25 +34,25 @@ import S from 'string'
 
 const options = {
   valueNames: ['id', 'text', 'timestamp', 'fromNow', 'time', 'tags', 'pinned'],
-  item: '<li class="col-xs-12 col-sm-6 col-md-6 col-lg-4">' +
-            '<span class="id" style="display:none;"></span>' +
-            '<a href="#">' +
-                '<div class="item">' +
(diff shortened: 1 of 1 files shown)


SISheogorath:
Nice work, thanks :) Your help is very welcome.

To be able to merge this, you would need to [sign-off your commit](https://github.com/hackmdio/hackmd/blob/master/CONTRIBUTING.md#sign-your-work). I'll test it later and merge it :)

jekrb:
Thank you!  I re-did my commit message with the sign-off!

ccoenen:
redoing the quotes inside the html block is a little overkill, in my opinion? I find double quotes to be way more common in HTML. At that point, it is not JavaScript (therefore JS code-styles should not apply). The surrounding quotes should be changed, I'm not against this PR in general.

(I'm speaking out in this case because I also often see people introducing camelCase in CSS class names, where dashes would be way more appropriate and things like that.)

jekrb:
@ccoenen thanks for the comment! I re-did this commit to use double-quotes.

PR #881, opened 2018-07-03
Repository is not renamed to docker-codimd yet

indigo423:
The repository is still named docker-hackmd instead docker-codimd.


SISheogorath:
I'll fix this as soon as the repository is renamed. Should become `codimd-docker`.

Issue #888, opened 2018-07-04
Speaker Mode is broken


NeoLegends:
Hey!

I used to make presentations with HackMD but now it seems like Revealjs's speaker mode (press s in slide mode) is broken and doesn't show anything.

Would love to see a quick fix!

NeoLegends:
Seems to be related to a CSP error. Not sure why it appears, though. There is no CSP rule in the notes.html, neither is there a manifest for hackmd.io.

```Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' vimeo.com https://gist.github.com www.slideshare.net https://query.yahooapis.com 'unsafe-eval' https://cdnjs.cloudflare.com https://cdn.mathjax.org https://www.google.com https://apis.google.com https://docs.google.com https://www.dropbox.com https://*.disqus.com https://*.disquscdn.com https://www.google-analytics.com https://stats.g.doubleclick.net https://secure.quantserve.com https://rules.quantcount.com https://pixel.quantserve.com https://js.driftt.com https://www.googletagmanager.com https://cdn.ravenjs.com 'nonce-3ebfb86a-ba2f-424b-8b2d-295299d7e8f8' 'sha256-EtvSSxRwce5cLeFBZbvZvDrTiRoyoXbWWwvEVciM5Ag=' 'sha256-NZb7w9GYJNUrMEidK01d3/DEtYztrtnXC/dQw7agdY4='". Either the 'unsafe-inline' keyword, a hash ('sha256-L0TsyAQLAc0koby5DCbFAwFfRs9ZxesA+4xg0QDSrdI='), or a nonce ('nonce-...') is required to enable inline execution.```

SISheogorath:
Thanks for the report!

This is caused by our CSP. I'm not sure if you run a self-hosted instance, on our demo instance https://hackmd-ce.herokuapp.com (not recommended) or on hackmd.io.

If it's hackmd.io, please open an issue in https://github.com/hackmdio/hackmd-io-issues/issues

On a self-hosted instance I would suggest to temporarily disable the CSPs.

SISheogorath:
I provided a fix upstream, let's see if it makes it. Otherwise we may need to maintain a fork. 

/cc @jackycute

jackycute:
I fixed it by adding the hash `sha256-L0TsyAQLAc0koby5DCbFAwFfRs9ZxesA+4xg0QDSrdI=` into our CSP scriptSrc directives.

[...]

PR #890, opened 2018-07-05
Add script src hash for speaker note to CSP directives

jackycute:
See #888

### DIFF
--- a/lib/csp.js
+++ b/lib/csp.js
@@ -71,7 +71,7 @@ function addInlineScriptExceptions (directives) {
   directives.scriptSrc.push(getCspNonce)
   // TODO: This is the SHA-256 hash of the inline script in build/reveal.js/plugins/notes/notes.html
   // Any more clean solution appreciated.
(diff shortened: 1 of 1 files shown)


lib/csp.js
```diff
@@ -72,6 +72,7 @@ function addInlineScriptExceptions (directives) {
   // TODO: This is the SHA-256 hash of the inline script in build/reveal.js/plugins/notes/notes.html
   // Any more clean solution appreciated.
   directives.scriptSrc.push('\'sha256-EtvSSxRwce5cLeFBZbvZvDrTiRoyoXbWWwvEVciM5Ag=\'')
+  directives.scriptSrc.push('\'sha256-L0TsyAQLAc0koby5DCbFAwFfRs9ZxesA+4xg0QDSrdI=\'')
```
SISheogorath: Shouldn't this one replace the upper one?

jackycute: Yeah, maybe. Sorry I couldn't remember it.
And seems it affect by when upgrade the reveal.js if it comes with changes in `notes.html`

SISheogorath:
Nice, just checked works with the new hash :+1:

Could you rebase your commits into one? `git rebase -i master` and mark the second one as fixup should do the job :)

PR #896, opened 2018-07-09
Upload images to the filesystem by default, rather than to imgur

maahl:
Hi,

It's easy to fail to realize that images are uploaded to imgur by default (I just got bitten by this), and this can be an issue in some situations.
I wrote this trivial patch to default to a filesystem upload, in case others think the same.
Not sure if I should bump the version number though, let me know if it's the case!

I'll update the docs where needed if this PR gets the green light.


SISheogorath:
Since we out default upload type is also the filesystem, I agree with you.

https://github.com/hackmdio/codimd/blob/d327bed65361372760af26354fec189e02b9dd4f/lib/config/default.js#L62

PR #902, opened 2018-07-23
Support 'host' & 'path' config options

ahihi:
hi,

this patch adds two new config options to specify where the server listens (along with corresponding environment variables and documentation):

- `host`: specifies the host on which to listen, allowing restriction to a particular IP
- `path`: specifies the path to a unix domain socket on which to listen (takes precedence over `host`/`port`)

### DIFF
--- a/app.js
+++ b/app.js
@@ -205,11 +205,21 @@ io.sockets.on('connection', realtime.connection)
 
 // listen
 function startListen () {
-  server.listen(config.port, function () {
+  var address
+  var listenCallback = function () {
     var schema = config.useSSL ? 'HTTPS' : 'HTTP'
-    logger.info('%s Server listening at port %d', schema, config.port)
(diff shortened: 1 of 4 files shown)


app.js
```diff
@@ -205,9 +205,19 @@ io.sockets.on('connection', realtime.connection)
 
 // listen
 function startListen () {
-  server.listen(config.port, function () {
+  var host = config.host || '0.0.0.0'
```
SISheogorath: Can we move defaults to the config section where we usually handle defaults?

https://github.com/hackmdio/codimd/blob/master/lib/config/default.js

ahihi:
hmm i thought i did already? or do you want me to squash as well?

SISheogorath:
Yes, please squash them. Since it's not needed to make 3 commits for this feature. As they are mostly fixups

ahihi:
ok :)

PR #910, opened 2018-07-31
Update to replace Twitter and Facebook to Riot

jackycute:
Replace the index page bottom social links of Twitter and Facebook to Riot chat.
To avoid making people confuse and let them ask in the place with better support.


SISheogorath:
:+1: Makes sense

May I'll find a nicer way to implement it later, but for now it should be fine ^^

Issue #916, opened 2018-08-12
Mattermost authentication doesn't work with Mattermost v5.0 and up


maxer456:
Hello,

Right now, Mattermost authentication is broken for Mattermost versions 5.0 and higher, trying to log in results in an "Internal Server Error" page.

The library CodiMD uses for interaction with the Mattermost API uses the older v3 API, which has been deprecated and removed. The library is no longer maintained and their [readme](https://github.com/mattermost/mattermost-driver-javascript) recommends using the v4 API driver from the [mattermost-redux](https://github.com/mattermost/mattermost-redux) repository.

Mattermost's [OAuth2 client sample](https://github.com/mattermost/mattermost-oauth2-client-sample) already uses the v4 API driver.

It would be awesome if this got fixed, if possible.

SISheogorath:
Can you try to use the [generic oauth module](https://github.com/hackmdio/codimd/pull/784/files#diff-b244c067c02a98d0b69bc198233055fd) (which I just realized is not documented). And if it works, would you like to contribute a guide ([like for GitHub](https://github.com/hackmdio/codimd/blob/master/docs/guides/auth/github.md)) on how to set it up?

Since I already noticed that the mattermost integration is outdated and I want to replace it in future with a guide instead.

maxer456:
Okay, I managed to get the generic OAuth module working with Mattermost. I have a questions though. Is the baseURL setting actually used somewhere? For authentication, I have to specify the full URL in every setting anyway.

I'll be happy to make a guide on how to set this up once I'm sure I have set it up correctly.

SISheogorath:
Fixed by #929

PR #929, opened 2018-08-28
Add an auth provider guide for Mattermost

maxer456:
The guide uses the generic OAuth2 module, as suggested in #916.

I couldn't find the baseURL setting used anywhere and a full URL needs to be specified in all the other URL settings, so it looks a bit redundant in the guide. Was the baseURL supposed to be prepended to the other URL settings?

Other than that, this setup works in my limited testing.

Signed-off-by: Ondřej Slabý <kron258@gmail.com>


SISheogorath:
Sounds good!

@indigo423 You may want to test that on your instance?

SISheogorath:
Thanks for your work, @maxer456! Feel free to join us at [`#codimd:matrix.org`](https://riot.im/app/#/room/#codimd:matrix.org)

PR #932, opened 2018-09-02
Upgrade to Webpack 4

davidmehren:
This PR contains (hopefully) everything needed to upgrade CodiMD to Webpack 4.
Beside upgrades to all the webpack-related dependencies and loaders, I also replaced the old `es2015` babel preset with the new `env` preset and moved the settings to `.babelrc`.

In my testing, everything looked and worked correctly, but I'm not sure I had a look at every feature.


davidmehren:
Okay, a few things I noticed just now:
- Travis fails on Node 7, because that version is not supported by `css-loader`. Not surprising, because Node 7 is EOL since more than a year.
- In https://github.com/hackmdio/codimd/blob/master/webpackBaseConfig.js#L355 the property `baseUrl` is set. Webpack 4 does not expect this property and refuses to run unless it is removed. The property is referenced in a few ejs templates (like `public/views/scripts.ejs`), obviously to set a base path. The weird thing is, that `baseUrl` is set to `<%- url %>` which is a ejs template string and I have no idea there the variable `url` comes from. I propose to just replace all of that with the `publicPath` property, that seems to be designed for just that: https://webpack.js.org/guides/public-path/
The remaining question is how to get the relative path from `config.json`. For now I removed the `baseUrl` property, which probably breaks subfolder depolyments.
- We should think about using https://github.com/survivejs/webpack-merge instead of `Object.assign` to merge the webpack configs,

SISheogorath:
Awesome work! I just checked and can't explain where it got the url setting, either. I supposed it got somewhere from the `config` module: 

https://github.com/hackmdio/codimd/blob/72894d1b7d816bd3c0f9197b7bf0674944f39901/lib/config/index.js#L73-L88

But no idea how it gets in there. Maybe just *magic*.

For subdirectory deployments I'm not sure at all. Right now, we already break them in all pre-build setup scripts. So I consider it more as a nice-to-have than a must-have.

---

Since this is working so far, I guess doing the switch to webpack-merge is probably something we can do as a second step?

I'll try to release 1.2.1 as soon as I can and then we can merge this to master and get towards 1.3.0 which should have this kind of upgrade as well as some others.

SISheogorath:
Can you rebase your changes to latest master, which no longer tests node 7?

SISheogorath:
Very awesome. Current ETA for 1.2.1 is 21st - 23rd of September.

We'll merge it afterwards as one of the first things!

PR #937, opened 2018-09-05
Add warning when using HTTP

SISheogorath:
We want to spread some awareness about plaintext HTTP connections. So we
show a big warning for plain HTTP connections and ask users to ask their
admin to setup HTTPS.

We may want to extent this to use link to a guide on how to do this.

Signed-off-by: Sheogorath <sheogorath@shivering-isles.com>

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -111,5 +111,6 @@
 	"Do you really want to delete your user account?": "Do you really want to delete your user account?",
 	"This will delete your account, all notes that are owned by you and remove all references to your account from other notes.": "This will delete your account, all notes that are owned by you and remove all references to your account from other notes.",
 	"Delete user": "Delete user",
-	"Export user data": "Export user data"
+	"Export user data": "Export user data",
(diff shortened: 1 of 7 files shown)


ccoenen:
is there a chance that this hidden text in the page will be picked up by robots? If so, I'd like to request a way that only injects it when needed. This wouldn't be a problem, as the check (and the whole client) only works with activated JavaScript anyway.

SISheogorath:
For search engines, robots should ignore it, since it's invisible. (Like they do with content in `<noscript>`-tags and similar

For other robots, yes, they may pick it up, but I see no risk when they do this. You can't use it to identify non-https instance, since all instances have it.

But yes, we can probably move this to be completely JS based so it doesn't show up at all without JS.

SISheogorath:
Yesterday I noticed that in the current way of implementation, we can easily translate it. The full JS implementation will lack of this. Currently thinking about how to implement a way to add translations to our JS sources.

PR #943, opened 2018-09-06
Some minor improvements for setup script

SISheogorath:
Since we use `yarn` for our container setup and try to enforce
dependencies, we should also use yarn in the setup script.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "dev": "webpack --config webpack.config.js --progress --colors --watch",
     "build": "webpack --config webpack.production.js --progress --colors --bail",
     "postinstall": "bin/heroku",
-    "start": "node app.js",
+    "start": "sequelize db:migrate && node app.js",
     "doctoc": "doctoc --title='# Table of Contents' README.md"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -11,7 +11,7 @@
     "dev": "webpack --config webpack.config.js --progress --colors --watch",
     "build": "webpack --config webpack.production.js --progress --colors --bail",
     "postinstall": "bin/heroku",
-    "start": "node app.js",
+    "start": "npx sequelize db:migrate && node app.js",
```
SISheogorath: @ccoenen With this change, would you prefer to publish this change in `1.2.1` or `1.3.0`?

ccoenen: we don't mention npx anywhere as far as I know. How is this more stable than node node_modules/.bin/sequelize (I presume, I haven't tested it). Does it warrant mandating npx?

I _think_ I'd rather see this in 1.3.

SISheogorath: I just checked, we can actually scratch the `npx` and simply use sequelize directly. It'll use the normal node way to resolve the the sequelize-cli instead of, how I just checked `npx` which may even installs it, when it can't find it.

PR #947, opened 2018-09-10
removing superfluous config parameters for template files

ccoenen:
All those pages still work, but now we're actually making use of https://github.com/hackmdio/codimd/blob/0017ddd3101f7c9445a70e8b38bf0046a0730f16/app.js#L170

### DIFF
--- a/app.js
+++ b/app.js
@@ -167,7 +167,7 @@ app.use(require('./lib/web/middleware/codiMDVersion'))
 
 // routes need sessions
 // template files
-app.set('views', path.join(__dirname, '/public/views'))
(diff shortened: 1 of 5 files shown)


SISheogorath:
We shouldn't use the `app.set` statement here. Or at least, make it configurable by config like we do it with the current view elements instead of having it statically using `__dirname` inside the code. Otherwise I agree that this makes it a lot simplier to render things :)

ccoenen:
commit is already contained in #940 (rebased https://github.com/hackmdio/codimd/pull/940/commits/bb80bc2292d2841776e8bfd6285f96194d990284)

PR #966, opened 2018-09-26
Fix document length limit on post

SISheogorath:
We recently introduced a new way to create notes using a post request
to the `/new` endpoint. This is not limited in size, other than pasting
a note in the editor. This patch should enforce this limit also on this
way.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -32,6 +32,9 @@ var response = {
   errorBadRequest: function (res) {
     responseError(res, '400', 'Bad Request', 'something not right.')
   },
+  errorTooLong: function (res) {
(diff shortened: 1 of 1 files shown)


ccoenen:
I think this check would be better suited for the Note-model, instead of fixing every endpoint. This might just be a sequelize validation that is run on every save or update.

SISheogorath:
I'm worried that when we add the validation to the note right now, we may cause some unhandled rejections that cause trouble. So I think I'll file a second PR for 1.3.0 which fixes that. Since I don't want to figure out all places, where a write to note could fail and crash something within the next 3 days.

PR #974, opened 2018-09-28
fix image paths

mcnesium:


SISheogorath:
Could you please fix the DCO error and make the commit message a bit more meaningful?

1. For that pull or clone your fork
2. change to your branch (`git checkout patch-1`)
3. Run `git commit --signoff --amend` and extend your commit message with a bit more meaning full content (for example where you fixed the image paths)
4. Push it to your existing branch and this way to your PR: `git push -f origin patch-1`

PR #978, opened 2018-10-02
Add asciinema integration

SISheogorath:
This PR is a working implementation for an asciinema integration. But due to security concerns it should be rewritten to not dynamically embed external JS from https://asciinema.org.

Instead we should try to rewrite this to use [asciinema-player](https://github.com/asciinema/asciinema-player) instead which is also [available on CDNJS](https://cdnjs.com/libraries/asciinema-player).

Feel free to pick this up and figure it out during Hacktober :)

### DIFF
--- a/lib/csp.js
+++ b/lib/csp.js
@@ -5,7 +5,7 @@ var CspStrategy = {}
 
 var defaultDirectives = {
   defaultSrc: ['\'self\''],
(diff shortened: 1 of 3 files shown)


ccoenen:
perhaps @alecdwm or @jreinert would give this a spin?

PR #984, opened 2018-10-04
Fix some webpack issues

SISheogorath:
We should use hashes for production builds. This prevents problems
during upgrades. At the same time, disable various optimizations and
features of webpack during development builds to speedup development
builds.

/cc @davidmehren

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "test": "npm run-script standard && npm run-script jsonlint",
     "jsonlint": "find . -not -path './node_modules/*' -type f -name '*.json' -o -type f -name '*.json.example' | while read json; do echo $json ; jq . $json; done",
     "standard": "node ./node_modules/standard/bin/cmd.js",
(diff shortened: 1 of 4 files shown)


SISheogorath:
Then let's undo the caching time change and merge this for now ^^

davidmehren:
Okay, so while cleaning up the webpack config I stumbled upon https://github.com/webpack/webpack-dev-server/issues/377#issuecomment-241258405.

It seems that using hashes in the dev-config is a bad idea and we should only use them in production.

SISheogorath:
Well, as far as I can say, we don't use `webpack-dev-server`. In our case it simply bloats the disk a bit.

The stated problem is related to the point that `webpack-dev-server` seems to cache the resulting files in memory without keeping track of where the files are originating from.

SISheogorath:
Completely refactored this PR. Please recheck @davidmehren

SISheogorath:
replaced by #1007

PR #987, opened 2018-10-04
How to use Nextcloud as OAuth2 Provider for CodiMD

ccoenen:
nextcloud and CodiMD play together very nicely! Let's tell the world about it!


pferreir:
Cool!

PR #990, opened 2018-10-04
Make oauth2 provider name accessible

SISheogorath:
Right now the feature exists but is almost not usable since the only way
to configure it is to know that it exists from reading the source code
and add it to config.json. This patch provides all needed changes so it
can be used by everyone including documentation.


ccoenen:
what does it do? I couldn't find any other references in the code? I also didn't notice any changes when I applied this to my freshly nextcloud-enabled CodiMD instance.

SISheogorath:
It changes the String in the button on the sign-in modal.

And if you didn't run this particular branch and used the environment variables, it doesn't work. Since right now (1.2.1) we can only use it from the config.json.

ccoenen:
Yup, I tried `CMD_` env vars. That explains it.

PR #991, opened 2018-10-05
Add full version string (and no AGPL violation detection)

SISheogorath:
Currently we only provide the version from `package.json`. This means
that during updates of instances, e.g. the demo instance, which runs
latest master instead of a stable release, changes are not reflected to
the webclient.

This patch adds a fullversion string that contains the current commit
and this way makes that clients are notified about changes.

Also adds an automated detection for AGPL clear violations and asks the user for fixing it in production environments.

### DIFF
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -8,6 +8,7 @@ const {merge} = require('lodash')
 const deepFreeze = require('deep-freeze')
 const {Environment, Permission} = require('./enum')
 const logger = require('../logger')
+const {getGitCommit, getGitHubURL} = require('./utils')
 
 const appRootPath = path.resolve(__dirname, '../../')
 const env = process.env.NODE_ENV || Environment.development
@@ -16,11 +17,17 @@ const debugConfig = {
 }
 
 // Get version string from package.json
-const {version} = require(path.join(appRootPath, 'package.json'))
+const {version, repository} = require(path.join(appRootPath, 'package.json'))
+
+const commitID = getGitCommit(appRootPath)
+const sourceURL = getGitHubURL(repository.url, commitID || version)
+const fullversion = commitID ? `${version}-${commitID}` : version
 
 const packageConfig = {
   version: version,
-  minimumCompatibleVersion: '0.5.0'
+  minimumCompatibleVersion: '0.5.0',
+  fullversion: fullversion,
+  sourceURL: sourceURL
 }
 
 const configFilePath = path.resolve(appRootPath, process.env.CMD_CONFIG_FILE ||
--- a/lib/config/utils.js
+++ b/lib/config/utils.js
@@ -1,5 +1,8 @@
 'use strict'
 
+const fs = require('fs')
+const path = require('path')
+
 exports.toBooleanConfig = function toBooleanConfig (configValue) {
   if (configValue && typeof configValue === 'string') {
     return (configValue === 'true')
@@ -19,4 +22,34 @@ exports.toIntegerConfig = function toIntegerConfig (configValue) {
     return parseInt(configValue)
   }
   return configValue
+}
+
+exports.getGitCommit = function getGitCommit (repodir) {
+  if (!fs.existsSync(repodir + '/.git/HEAD')) {
+    return undefined
+  }
+  let reference = fs.readFileSync(repodir + '/.git/HEAD', 'utf8')
+  if (reference.startsWith('ref: ')) {
+    reference = reference.substr(5).replace('\n', '')
+    reference = fs.readFileSync(path.resolve(repodir + '/.git', reference), 'utf8')
+  }
+  reference = reference.substr(5).replace('\n', '')
+  return reference
+}
+
+exports.getGitHubURL = function getGitHubURL (repo, reference) {
+  // if it's not a github reference, we handle handle that anyway
+  if (!repo.startsWith('https://github.com') && !repo.startsWith('git@github.com')) {
+    return repo
+  }
+  if (repo.startsWith('git@github.com') || repo.startsWith('ssh://git@github.com')) {
+    repo = repo.replace(/^(ssh:\/\/)?git@github.com:/, 'https://github.com/')
+  }
+
+  if (repo.endsWith('.git')) {
+    repo = repo.replace(/\.git$/, '/')
+  } else if (!repo.endsWith('/')) {
+    repo = repo + '/'
+  }
+  return repo + 'tree/' + reference
 }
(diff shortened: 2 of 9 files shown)


lib/config/utils.js
```diff
@@ -20,3 +23,23 @@ exports.toIntegerConfig = function toIntegerConfig (configValue) {
   }
   return configValue
 }
+
+exports.getGitCommit = function getGitCommit (repodir) {
+  if (!fs.existsSync(repodir + '/.git/HEAD')) {
+    return undefined
+  }
+  let reference = fs.readFileSync(repodir + '.git/HEAD', 'utf8')
```
azyobuzin: A slash is required before `'.git/HEAD'` because `repodir` may not be ended with a slash.

lib/config/index.js
```diff
@@ -18,9 +19,15 @@ const debugConfig = {
 // Get version string from package.json
 const {version} = require(path.join(appRootPath, 'package.json'))
 
+const commitID = getGitCommit(appRootPath)
+const sourceURL = getGitHubURL(commitID || version)
```
azyobuzin: `repo` paramter is not specified. In the current implementation, `sourceURL` will be `CommitID/tree/undefined`.

SISheogorath: That's correct, I already noticed that and currently think about the best way to get the real upstream repo for a branch out of the existing repo settings without re-build the whole git tooling on this

lib/config/utils.js
SISheogorath: Oh I missed that one ^^

SISheogorath:
New version fixes the GitHub URL thing and adds AGPL violation test. This should help to make people aware of how AGPL works.

lib/config/utils.js
```diff
@@ -20,3 +23,33 @@ exports.toIntegerConfig = function toIntegerConfig (configValue) {
   }
   return configValue
 }
+
+exports.getGitCommit = function getGitCommit (repodir) {
+  if (!fs.existsSync(repodir + '/.git/HEAD')) {
```
ccoenen: to get rid of this error or warning, i just need to keep my `.git` directory outside the project directory?
https://stackoverflow.com/a/505517/2413043

lib/config/index.js
```diff
@@ -113,6 +121,20 @@ if (config.gitlab && config.gitlab.version !== 'v4' && config.gitlab.version !==
 // Only update i18n files in development setups
 config.updateI18nFiles = (env === Environment.development)
 
+// Verify that sourceURL is correct
+if (env === Environment.production) {
+  request.get(config.sourceURL)
```
ccoenen: Just to have this on record: I'm not a fan of pinging home or anywhere on server startup. Not from a technological standpoint but more so not from a surveillance standpoint. I don't need to tell Github/Microsoft and the US Government where my server is located every time CodiMD comes up.

lib/config/utils.js
SISheogorath: Yes. You can also do a lot of other things. (as mentioned on Matrix). And as mentioned there, the idea is not to prevent people from breaking AGPL but giving them a heads up, when something seems to be not right.

lib/config/index.js
SISheogorath: That's right. That's why people are free to set the sourceURL. The reason to verify the source URL is that we have a plenty of forks out there and a lot of them have modifications. Most of them still don't change any data about the repository location or similar.

This PR adds a link to the entry page which points to the source code. It makes no sense to add this link when the link is wrong (which it would be, for all these mentioned forks). So it would mislead the user to a source code that isn't actually the source code of the instance they are using.

So I'm open for alternative solutions for that problem, but the idea is to catch forks, that are unaware that they have to change something to get this link right. It doesn't protect from people who want to fool users intentionally, but at least unintentional fooling should be prevented.

As mentioned in Matrix. One of the alternative thoughts I had was zipping the CodiMD directory. But that again is dangerous because someone may keeps backups of the config.json, places private keys next to the server, … and this way would cause a high risk for deployments. So the alternative is to link to the correct source and the probably worst UX is when you end up in a dead end. And in this case it has [legal implications](https://github.com/hackmdio/codimd/blob/32af96aa376576d060e16a2c0ff7ed5719d24e91/LICENSE#L542-L551).

As AGPL states you not only have to make your users aware that they CAN recieve a copy of the source code (which we do with the link), you also have to provide it in a way that it can be recieved from a server at no charge. Granted, this could also mean you send out an email to them, but I think this is quite uncommon these days and probably up to legal interpretation. Still I think misleading/wrong links are worse than pinging GitHub. (Which is also avoidable when you set the link to an own page that provides the source code.)

ccoenen:
We were debating this at length on matrix, but my main issue with this is, I do not believe it serves any purpose.

let's say we have to kinds of people:

- *cooperative* Those that care about doing "the right" thing (let's say those are 90%, I'd argue its closer to 100%, but let's say it's just 90%)
- *mischievous* Those that will do anything for their personal advantage (the rest)

And we have two kinds of situations:

- *vanilla* those that run vanilla releases (let's say this is 75% of installations)
- *modified* those that change the code in any way. (let's say this is the other 25%)

Of those groups, we don't need to care about 

- cooperative / vanilla and mischievous / vanilla -> combined 75%

Both could just be a static link to our github repository. No Changes? Nothing to do here. We also don't need to care about

mischievous / modified -> 5%

because they will just circumvent any kind of "protection" we come up with. We gain nothing by any means we're adding in here. So let's talk about

- cooperative / modified -> 20%

Of those, a large number will probably honor the AGPL already. They may be contributors to upstream or they may be giving out source code on request of their users (which is perfectly fine with the AGPL!)

So what do we do with the rest, that is _at worst_ 10% of installations? Those people that have good intentions and may *potentially* be breaking AGPL by accident. We're breaking their servers and slap "YOU'RE DOING IT WRONG* in their faces. While they were not even technically really breaking the AGPL! We're not entitled to their changes. Their USERS are entitled to these changes. We know nothing about how they may or may not be distributing their code to their users. So we're also getting false positives here.

The cooperative but accidentally, potentially breaking the AGPL is better served by a friendly notice about what would be a good idea ("we'd encourage you to make your source code available via Github/Gitlab or zipfiles/tarballs").

At the same time we're adding some kind of phone home mechanism that I really don't like. And we're delivering this to all installations. In my opinion this just recreates the mistakes music and gaming industries were/are doing with copy protections. In the end, we're creating more problems for well-intentioned people than we're solving.

I propose to keep sourceURL in the config. We're adding this to the setup instructions: "if you modify the code, your users have a right to receive those code changes, which is part of our License, the AGPL". Done. Nothing phones home. Nothing breaks if they don't do it.

SISheogorath:
Since we didn't came up with a better solution, I removed the validation part and instances with a non correct source URL simply end up with a 404. That's not really nice, but since we don't want to phone anywhere and there is no other way to validate, that's how it is.

If there are no further requests for changes I'll merge this version tomorrow.

ccoenen:
the config parameters are not yet included in the `README`

PR #994, opened 2018-10-06
Update supported node.js version

xf-:
Support includes v9.x, but no v10.x

Console message:
```
error CodiMD@1.2.1: The engine "node" is incompatible with this module. Expected version ">=6.x <10.x". Got "10.11.0"
```
https://github.com/nodejs/Release
Node.js LTS 10.x starts in October and i saw no PR to change compatible version

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ Just to more confusion: We are still friends with HackMD :heart:
 
 ### Prerequisite
 
-- Node.js 6.x or up (test up to 7.5.0)
+- Node.js 6.x or up (test up to 7.5.0) and <10.x
 - Database (PostgreSQL, MySQL, MariaDB, SQLite, MSSQL) use charset `utf8`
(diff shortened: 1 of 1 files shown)


SISheogorath:
We explicitly set the version <10 because version 10 broke scrypt (and iirc also uws which is replaced now).

Since scrypt didn't get an update since 2 years, I don't expect that to be fixed magically, when node didn't suddenly decide to revert their API changes.

xf-:
i added it only as information in README - no code - next time to use v8 or v9 from beginning

SISheogorath:
Oh just saw that you edit the readme, thought you edit the package.json.

We are aware of node-scrypt. It just needed some work [(which now waits for a PR)](https://matrix.to/#/!ybaSlrrMNLeuoQuUuR:matrix.org/$1537117462311625kDQXW:matrix.org) because node-scrypt and the scrypt package we used, are not compatible in their output formats.

---

Apart from that, please fix the DCO and we can merge your PR :)

xf-:
ok - i just patched it via web edit foo

PR #1002, opened 2018-10-09
Fix #1001: get only project user is member of (and return max of results)

### LINKED ISSUE #1001: [Gitlab snippets] Too many projects to choose from
When listing all projects through gitlab API (v4), all projects are returned, and the results are only the 20 first.

If the codimd user is not a member of those projects, no project can be selected to export snippets to.

This is because the `/projects` uri changed contract between v3 and v4 of gitlab api. It used to return only project the user was a member of.

micedre:


SISheogorath:
Are you around in the [Matrix community](https://matrix.to/#/#codimd:matrix.org)? If not, would be nice to have you around ^^

Thanks for your work!

PR #1003, opened 2018-10-09
Integrate forkawesome with CodiMD for easier usage

SISheogorath:
See commit comments for details.

Basically adds `{%fa <icon name> [<color>]%}` to use the icons instead of weird `<i>` statements.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "express-session": "^1.14.2",
     "file-saver": "^1.3.3",
     "flowchart.js": "^1.6.4",
-    "font-awesome": "^4.7.0",
+    "fork-awesome": "^1.1.3",
     "formidable": "^1.0.17",
     "gist-embed": "~2.6.0",
     "graceful-fs": "^4.1.11",
--- a/public/docs/features.md
+++ b/public/docs/features.md
@@ -127,6 +127,12 @@ Use the syntax `[TOC]` to embed table of content into your note.
 You can type any emoji like this :smile: :smiley: :cry: :wink:
 > See full emoji list [here](http://www.emoji-cheat-sheet.com/).
 
+## Icons
+You can use any symbol from [ForkAwesome](https://forkawesome.github.io/). To use them you can either use their icon classes. As alternative use our integration: `{%fa <icon name> <icon color>%}`.
+
+{%fa rss%} {%fa lightbulb-o orange%} {%fa nextcloud #0082C9%}
+<i class="fa fa-rss" aria-hidden="true"></i> <i class="fa fa-lightbulb-o" aria-hidden="true" style="color: orange;"></i> <i class="fa fa-nextcloud" aria-hidden="true" style="color: #0082C9;"></i>
+
 ## ToDo List:
 - [ ] ToDos
   - [x] Buy some salad
--- a/public/js/extra.js
+++ b/public/js/extra.js
@@ -1056,6 +1056,22 @@ md.renderer.rules.fence = (tokens, idx, options, env, self) => {
 /* Defined regex markdown it plugins */
 import Plugin from 'markdown-it-regexp'
 
+const forkawesomePlugin = new Plugin(
+    // regexp to match
+    /{%fa\s*([\d\D]*?)(\s([\d\D]*?))?\s*%}/,
+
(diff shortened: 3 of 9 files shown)


ccoenen:
Is there a precedence for this syntax? I haven't seen this before.

SISheogorath:
I took it from the [externals](https://github.com/hackmdio/codimd/blob/e0e037b5e1c30968322b0fc53274b9a9b052e593/public/docs/features.md#externals) we currently have. Wasn't sure about any other syntax that doesn't conflict with our existing integrations and possible future ones. But I'm open for suggestions.

ccoenen:
ah, right :-)

SISheogorath:
Thought about integrating it with the way of referencing smileys `:something:` but couldn't figure out a way to decide what is a smiley and what is an icon ^^

SISheogorath:
@almereyda hint: Yesterday I already merged https://github.com/hackmdio/codimd/pull/1003/commits/5212bbf9c4bb14d0ee5f009512df0bb6f7a475d0 so you can already try forkAwesome icons on the demo instance. Just the further integration was put into an own PR to get forward.

almereyda:
Please note Mermaid (https://mermaidjs.github.io/flowchart.html - at the bottom: Basic support for fontawesome) also uses a similar notation.

SISheogorath:
@almereyda Huh, nice! may we should go for `fa:<icon name>` just seems to be a bit too easy so we could get a lot false positives :/

Any opinions on it?

ccoenen:
`fa:<name>` will go and haunt is, in my opinion. To be more similar, one _could_ use `{%fa:<icon name>%} ` But I don't think it's a good idea to have this completely without any markup.

ccoenen:
If someone were to write auto-complete with search for icons for this, that person would be my hero!

SISheogorath:
Here is the list of all icons:
https://github.com/ForkAwesome/Fork-Awesome/blob/next-release/src/icons/icons.yml

Now all we need to do is taking this and make it a JSON that we can integrate during the build process ^^

ccoenen:
this is all that's required to make this a nice simple JS object:

```js
const fs = require('fs')

const yaml = require('js-yaml')

const data = fs.readFileSync('./icons.yml')
const icons = yaml.safeLoad(data)
const sortedIcons = icons.icons.sort((a,b) => a.id.localeCompare(b.id))

const index = sortedIcons.map((i) => {
  i.filter = i.filter || []
  i.categories = i.categories || []
  i.aliases = i.aliases || []
  const ingredients = [i.name, i.id, ...i.filter, ...i.categories, ...i.aliases]
  return [i.id, ingredients.join(', ')]
})

console.log(index);
```

we're putting out an array of arrays where the first element is the id of the fork-awesome icon. Then were' mumbling up a long string of things we might want to search.

lookupHash now looks like:
```js
[
  [ '500px', '500px, 500px, Brand Icons' ],
  [ 'activitypub', 'ActivityPub, activitypub, Brand Icons' ],
  [ 'address-book', 'Address Book, address-book, Web Application Icons' ],
  [ 'address-book-o', 'Address Book Outlined, address-book-o, Web Application Icons' ],
  [ 'address-card', 'Address Card, address-card, Web Application Icons, vcard' ],
  // and so on
]
```

one could then simply search this with:

```js
const searchTerm = new RegExp('right', 'i') // yes I know /editor/i, but this might be from an input!
const result = index.filter((item) => searchTerm.test(item[1]))

console.log(result);
```

this will now output just the matching things:

```js
[
  [ 'align-right', 'align-right, align-right, text, Text Editor Icons' ],
  [ 'angle-double-right', 'Angle Double Right, angle-double-right, raquo, quote, next, forward, arrows, Directional Icons' ],
  [ 'angle-right', 'angle-right, angle-right, next, forward, arrow, Directional Icons' ],
  [ 'arrow-circle-o-right', 'Arrow Circle Outlined Right, arrow-circle-o-right, next, forward, Directional Icons' ],
  [ 'arrow-circle-right', 'Arrow Circle Right, arrow-circle-right, next, forward, Directional Icons' ],
  [ 'arrow-right', 'arrow-right, arrow-right, next, forward, Directional Icons' ],
  // and so on
]
```

PR #1004, opened 2018-10-10
Add autocomplete for highlight.js languages into codemirror

SISheogorath:
Right now we support code highlighting for rust, but it doesn't appear
in autocomplete of codemirror, because codemirror is not aware of it.

This patch lets highlightjs simply tell codemirror, what it supports and
adds this to the autocomplete list.

### DIFF
--- a/public/js/index.js
+++ b/public/js/index.js
@@ -17,6 +17,7 @@ import toMarkdown from 'to-markdown'
 import { saveAs } from 'file-saver'
 import randomColor from 'randomcolor'
 import store from 'store'
+import hljs from 'highlight.js'
 
 import _ from 'lodash'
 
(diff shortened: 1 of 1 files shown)


azyobuzin:
It seems that only some languages are added to the autocomplete list if `useCDN` is true.

screenshot

SISheogorath:
There are 29 languages we have as default and we add the the 23 from highlight,js to auto-complete.

So from this perspective it's fine.

azyobuzin:
highlight.js has 176 languages and they all are added if `useCDN` is false. So the behavior varies depending on `useCDN` setting. And the first goal, 'rust' is not included in the default 23 languages.

SISheogorath:
Mhm, right. So we basically uncovering a bug by integrate things well .-. Wonderful.

Seems like the fix for the bug is actually reducing the number of supported languages since we won't bring highlightjs to publish a `.full.min.js` version that contains all 176 languages. So the only other option would be to include the other remaining 153 languages one by one from the CDN and this will kill the loadtime, so I'm not about to do that.

ccoenen:
lazy loading the highlighters would make much more sense. Not sure if it can be done.

PR #1007, opened 2018-10-10
Webpack 4 refactor & docs

davidmehren:
This PR contains a major refactor of the webpack configs and adds documentation:
- We now use `webpack-merge` to merge common and prod/dev configs.
- Options that are already included in the webpack 4 defaults were deleted.
- In dev mode, "cheap" source maps are used.
- The config for the "html export" feature was split into its own file to avoid duplicated code.
- Development docs for the webpack config was added.


SISheogorath:
What a wonderful work. Thank you so much ^^

PR #1021, opened 2018-10-18
Webpack: Cleanup common config

davidmehren:
Just a small cleanup of the common webpack config.

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -392,7 +392,7 @@ module.exports = {
   module: {
     rules: [{
       test: /\.js$/,
-      use: [{loader: 'babel-loader'}],
+      use: [{ loader: 'babel-loader' }],
       exclude: [/node_modules/, /public\/vendor/]
     }, {
       test: /\.css$/,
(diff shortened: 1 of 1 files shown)


webpack.common.js
```diff
@@ -1,14 +1,9 @@
...
+const webpack = require('webpack')
+const path = require('path')
+const HtmlWebpackPlugin = require('html-webpack-plugin')
+const CopyWebpackPlugin = require('copy-webpack-plugin')
 const MiniCssExtractPlugin = require('mini-css-extract-plugin')
 
-// Fix possible nofile-issues
-var fs = require('fs')
-var gracefulFs = require('graceful-fs')
-gracefulFs.gracefulify(fs)
```
SISheogorath: I would like to keep this.

image

davidmehren: Ahh, that's what this is for ^^

Let's keep removing this out of this PR then. Do we have an open Issue to get to the bottom of the error?

SISheogorath: I don't think so.

https://github.com/systemd/systemd/pull/10244 might solves it soon, without touching any (Node)JS code :smile: 

Apart from that, I guess it's simply related to the size of our codebase :/

SISheogorath:
I like it :)

PR #1027, opened 2018-10-28
Add download action to published notes

asg017:
Right now, you can't perform the `/download` action on a published note.

Currently, with a normal note, `/download` works:

```
https://hackmd.io/B58Juygfrj4rg8AgdA/download    // ✔️ 
```

But, with the published, `download` isnt an allowed action:
```
https://hackmd.io/s/dtG9j/download         // ⛔️ 
```

This PR makes it so `download` works on published notes:
```
https://hackmd.io/s/dtG9j/download         //  ✔️
```

I think `download` should be an allowed action on published notes - published notes are read-only and downloading is a read-only operation - and it allows for a API-ish endpoint for easily downloading published notes.

🍕 

Signed-off-by: Alex Garcia <alexsebastian.garcia@gmail.com>

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -468,6 +468,8 @@ function publishNoteActions (req, res, next) {
   findNote(req, res, function (note) {
     var action = req.params.action
     switch (action) {
+      case 'download':
+        actionDownload(req, res, note)
       case 'edit':
         res.redirect(config.serverURL + '/' + (note.alias ? note.alias : models.Note.encodeNoteId(note.id)))
         break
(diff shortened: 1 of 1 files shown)


SISheogorath:
Nice to see your contribution!

Yes, it seems to be useful, but just as hint: hackmd.io is not running CodiMD and this way your changes won't affect notes there. (for details please check: https://github.com/hackmdio/codimd#hackmd-ce-became-codimd)

At the same time, you are may interested in the [CodiMD-CLI project](https://github.com/hackmdio/codimd-cli). It's very basic since an API isn't really existing yet, but this should change in future.

I'll leave your change open for a few more hours so everyone has a change to comment on it. Meanwhile, can you squash your commits together before we merge this PR?

Welcome to the community :)

PR #1030, opened 2018-10-29
add option reset in bin/manage_users

Eronana:
- add an option `reset` to reset user password
- refactored some code in ES6(eliminate var, use template string)
- refactored the way to check arguments
- check type of argv[action] to ensure that is `string` to avoid a pass a boolean email to database
- check type of pass

### DIFF
--- a/bin/manage_users
+++ b/bin/manage_users
@@ -22,6 +22,15 @@ Usage: bin/manage_users [--pass password] (--add | --del) user-email
 	process.exit(1);
 }
 
+function getPass(argv, action) {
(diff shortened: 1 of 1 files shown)


SISheogorath:
Nice to see your work :) Seems like you putted some work into it as you also refactored various parts.

From my perspective it would be nice to have the refactoring part and feature adding part split up into two commits.

This keeps the commit history a bit cleaner and makes it easier to revert changes for whatever reason in the future.

SISheogorath:
Thanks a lot!

PR #1040, opened 2018-11-06
Fix migration failure due to change on error messages

sunbit:
I think this already happened in the past, i guess is some library version change.


SISheogorath:
Can you please squash your commits together? Then I'll merge this one. :crossed_fingers: that there won't be a lot more versions of those errors.

PR #1041, opened 2018-11-07
Fix menu when gitlab is enabled

micedre:
When gitlab is enabled with scope API, the menu Export/Import Snippets is not displayed.
This is because gitlab var stay indefined in public/views/codimd/header.ejs.

The fix here is to add a var `enableGitlabSnippets` and then use it in header.ejs.

### DIFF
--- a/app.js
+++ b/app.js
@@ -198,6 +198,7 @@ app.locals.authProviders = {
   email: config.isEmailEnable,
   allowEmailRegister: config.allowEmailRegister
 }
+app.locals.enableGitlabSnippets = (!config.gitlab.scope || config.gitlab.scope === 'api')
 
 app.use(require('./lib/web/baseRouter'))
 app.use(require('./lib/web/statusRouter'))
--- a/app.js
+++ b/app.js
@@ -198,7 +198,11 @@ app.locals.authProviders = {
   email: config.isEmailEnable,
   allowEmailRegister: config.allowEmailRegister
 }
-app.locals.enableGitlabSnippets = (!config.gitlab.scope || config.gitlab.scope === 'api')
+
+// Export/Import menu items
+app.locals.enableDropBoxSave = config.isDropboxEnable
+app.locals.enableGitHubGist = config.isGitHubEnable
+app.locals.enableGitlabSnippets = config.isGitlabSnippetsEnable
 
 app.use(require('./lib/web/baseRouter'))
 app.use(require('./lib/web/statusRouter'))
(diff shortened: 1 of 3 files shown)


app.js
```diff
@@ -198,6 +198,7 @@ app.locals.authProviders = {
   email: config.isEmailEnable,
   allowEmailRegister: config.allowEmailRegister
 }
+app.locals.enableGitlabSnippets = (!config.gitlab.scope || config.gitlab.scope === 'api')
```
SISheogorath: Can we move the `(!config.gitlab.scope || config.gitlab.scope === 'api')`-part to [`./lib/config/index.js`](https://github.com/hackmdio/codimd/blob/master/lib/config/index.js)?

public/views/codimd/header.ejs
```diff
@@ -32,7 +32,7 @@
                 </li>
                 <li role="presentation"><a role="menuitem" class="ui-extra-slide" tabindex="-1" href="#" target="_blank" rel="noopener"><i class="fa fa-tv fa-fw"></i> <%= __('Slide Mode') %></a>
                 </li>
-                <% if((typeof github !== 'undefined' && github) || (typeof dropbox !== 'undefined' && dropbox) || (typeof gitlab !== 'undefined' && gitlab && (!gitlab.scope || gitlab.scope === 'api'))) { %>
+                <% if((typeof github !== 'undefined' && github) || (typeof dropbox !== 'undefined' && dropbox) || enableGitlabSnippets) { %>
```
SISheogorath: And can you fix github and dropbox along the line, since they'll definitely break as well.

micedre:
Done, is it what you had in mind?
I can't test github or dropbox in my environment (only that the menu are here).

public/views/codimd/header.ejs
```diff
@@ -134,16 +134,16 @@
                     </li>
                     <li role="presentation"><a role="menuitem" class="ui-extra-slide" tabindex="-1" href="#" target="_blank" rel="noopener"><i class="fa fa-tv fa-fw"></i> <%= __('Slide Mode') %></a>
                     </li>
-                    <% if((typeof github !== 'undefined' && github) || (typeof dropbox !== 'undefined' && dropbox) || (typeof gitlab !== 'undefined' && gitlab && (!gitlab.scope || gitlab.scope === 'api'))) { %>
+                    <% if(enableGitHubGist || enableDropBoxSave || enableGitlabSnippets ) { %>
```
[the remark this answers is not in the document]
micedre: sure !

SISheogorath:
Nice work!

PR #1048, opened 2018-11-10
Add an etherpad migration guide

dsprenkels:
I recently migrated from Etherpad to CodiMD and migrated the pads
that were still on Etherpad. On Matrix [it was mentioned](https://matrix.to/#/!ybaSlrrMNLeuoQuUuR:matrix.org/$1541666805641585LnWVS:matrix.org) that a migration guide for Etherpad would be a useful contribution.

In this guide I share how a migration can be done. I am not completely sure if the script that is included is
completely error-free. Readers/reviewers should be aware that there may be bugs.

### DIFF
--- a/docs/guides/migrate-etherpad.md
+++ b/docs/guides/migrate-etherpad.md
@@ -1,0 +1,131 @@
+Pad migration guide from etherpad-lite
+===
+
+The goal of this migration is to do a "dumb" import from all the pads in Etherpad, to notes in
+CodiMD. In particular, the url locations of the pads in Etherpad will be lost. Furthermore, any
+metadata in Etherpad, such as revisions, author data and also formatted text will not be migrated
+to CodiMD (only the plain text contents).
+
+Note that this guide is not really meant as a support guide. I migrated my own Etherpad to CodiMD,
+and it turned out to be quite easy in my opinion. In this guide I share my experience. Stuff may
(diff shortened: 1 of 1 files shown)


docs/guides/migrate-etherpad.md
```diff
@@ -0,0 +1,131 @@
...
+# This script uses the codimd command line script[1] to import a list of pads from 
+# [1]: https://github.com/hackmdio/codimd-cli/blob/master/bin/codimd
+
+# The base url to where etherpad is hosted
+ETHERPAD_SERVER="https://etherpad.example.com"
+
+# The base url where codimd is hosted
+CODIMD_SERVER="https://codimd.example.com"
+
+# Write a list of pads and the urls which they were migrated to
```
ccoenen: Great idea! 👍

ccoenen:
This is great! Very detailed. One more thing one could do after the import: The Notes Table in CodiMD allows for an `alias`. The CLI can't yet post to a new note with alias, but with the rewrites file, it would probably be possible to set aliases for each line afterwards.

SISheogorath:
@ccoenen the bigger problem is probably that the `/p/` URLs would lead to presentation mode instead of  the normal editor. So setting up the redirects on the reverse proxy actually makes sense from my point of view.

PR #1053, opened 2018-11-12
Disallow creation of robots.txt in freeurl

### LINKED ISSUE #1052: Make robots.txt a special note
Currently, in freely mode, `/robots.txt` is editable as if it were a regular note. However, `robots.txt` has a special meaning for web browsers. Because of this my CodiMD server awkwardly has a `robots.txt` [page](https://codimd.dsprenkels.com/robots.txt) that was created by a bot some time ago.

Can we treat the `/robots.txt` note as a special note and return a HTTP `204 No Content` or `404 Not Found` status code?

dsprenkels:
Add a configuration setting to "hard"-disable creation of notes as set by the configuration value `CMD_FORBIDDEN_NOTE_IDS`. This defaults to `['robots.txt', 'favicon.ico']`, because these files are often accidentally created by bots and browsers.

As of now, it does not seem that a custom `robots.txt` file is needed. Looking from the other code in the source tree the general convention seems to be using `x-robots-tag` headers. So we can just return a `404 Not Found` error on `/robots.txt`.

However, we don't want to break an installation that already has these notes created (purposefully). So instead of blocking requests to `/robots.txt`, we will block any request to create it.

This PR fixes #1052.

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,6 +186,7 @@ There are some config settings you need to change in the files below.
 | `CMD_ALLOW_ANONYMOUS` | `true` or `false` | set to allow anonymous usage (default is `true`) |
 | `CMD_ALLOW_ANONYMOUS_EDITS` | `true` or `false` | if `allowAnonymous` is `true`, allow users to select `freely` permission, allowing guests to edit existing notes (default is `false`) |
 | `CMD_ALLOW_FREEURL` | `true` or `false` | set to allow new note creation by accessing a nonexistent note URL |
+| `CMD_FORBIDDEN_NODE_IDS` | `'robots.txt'` | disallow creation of notes, even if `CMD_ALLOW_FREEURL` is `true` |
 | `CMD_DEFAULT_PERMISSION` | `freely`, `editable`, `limited`, `locked` or `private` | set notes default permission (only applied on signed users) |
 | `CMD_DB_URL` | `mysql://localhost:3306/database` | set the database URL |
 | `CMD_SESSION_SECRET` | no example | Secret used to sign the session cookie. If non is set, one will randomly generated on startup |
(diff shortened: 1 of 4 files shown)


SISheogorath:
We should also add `api` and probably everything that has just 1 letter to this.

I'm so sure there will be the point that we want some endpoint and at least 3 admins/users have setup a super important note there.

dsprenkels:
I'll add `api`, but I don't really see a point in adding all strings of one letter. They are not really specific.

> I'm so sure there will be the point that we want some endpoint and at least 3 admins/users have setup a super important note there.

Maybe add a note to the documentation on `freeurl` mode that this is risk they take?

README.md
```diff
@@ -186,6 +186,7 @@ There are some config settings you need to change in the files below.
 | `CMD_ALLOW_ANONYMOUS` | `true` or `false` | set to allow anonymous usage (default is `true`) |
 | `CMD_ALLOW_ANONYMOUS_EDITS` | `true` or `false` | if `allowAnonymous` is `true`, allow users to select `freely` permission, allowing guests to edit existing notes (default is `false`) |
 | `CMD_ALLOW_FREEURL` | `true` or `false` | set to allow new note creation by accessing a nonexistent note URL |
+| `CMD_FORBIDDEN_NODE_IDS` | `['robots.txt']` | disallow creation of notes, even if `CMD_ALLOW_FREEURL` is `true` |
```
ccoenen: this is different from the config.json explanation below. Reading the code, I believe this one to be the correct description.

README.md
```diff
@@ -284,6 +285,7 @@ There are some config settings you need to change in the files below.
 | `allowAnonymous` | `true` or `false` | set to allow anonymous usage (default is `true`) |
 | `allowAnonymousEdits` | `true` or `false` | if `allowAnonymous` is `true`: allow users to select `freely` permission, allowing guests to edit existing notes (default is `false`) |
 | `allowFreeURL` | `true` or `false` | set to allow new note creation by accessing a nonexistent note URL |
+| `forbiddenNoteIDs` | `true` or `false` | disallow creation of notes, even if `allowFreeUrl` is `true` |
```
ccoenen: different from the env var description (see above). Reading the code, I believe this one to be faulty.

dsprenkels: Yes, that's obviously a copy-paste error.

dsprenkels: Updated.

README.md
SISheogorath: This should be a comma separated list. no brackets.

SISheogorath: Yes. It's just the `toArrayConfig()`-function that doesn't use brackets.

PR #1055, opened 2018-11-13
Upgrade winston / refactor logging

SISheogorath:
Our log library got a new major version which should be implemented.

That's exactly what this patch does. Implementing the new version of the
logging library.

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -1,23 +1,19 @@
 'use strict'
-const winston = require('winston')
+const {createLogger, format, transports} = require('winston')
 
-class Logger extends winston.Logger {
-  // Implement stream.writable.write interface
-  write (chunk) {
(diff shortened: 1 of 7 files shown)


lib/logger.js
```diff
@@ -1,23 +1,19 @@
...
+    format.uncolorize(),
+    format.timestamp(),
+    format.align(),
+    format.splat(),
+    format.printf(info => `${info.timestamp} ${info.level}: ${info.message}`)
+  ),
   transports: [
-    new winston.transports.Console({
+    new transports.Console({
       level: 'debug',
```
SISheogorath: I would really love to get rid of the debug logging as default, but since we use this logging class in the config module, we shouldn't include the config here. So I'm not sure how to solve this.

PR #1057, opened 2018-11-14
switching to eslint for code checking

ccoenen:
Most rules degraded to WARN, so we don't go insane. ~~Over time, I'd like to reduce the amount of globals, if possible~~ (globals now only affect the client part of the code). I'd also like to reinstate the defaults instead of overriding as warning. But before we do that, we should probably discuss if standard is still the way to go.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,3 @@
+lib/ot
+public/vendor
+public/build
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,21 @@
+module.exports = {
+  "root": true,
+  "extends": "standard",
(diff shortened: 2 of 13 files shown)


ccoenen:
travis (correctly) called me out on an error before:
https://travis-ci.org/hackmdio/codimd/jobs/454986654

ccoenen:
now it works (despite the warnings):
https://travis-ci.org/hackmdio/codimd/jobs/455008623

I plan on adding `/public` back into the mix.

SISheogorath:
I like it :+1: will look that we leave it open for a day and merge it tomorrow :)

PR #1063, opened 2018-11-19
After removing ws, node version 10 should work

SISheogorath:
In my local environment I switched to Fedora 29. Fedora 29 comes with
NodeJS version 10.

As far as I can say, it works, so let's try to remove the restriction to
"<10.x"

**Edit:** We should also consider to provide 1.4.0 (yes, explicitly not 1.3.0) using Node version 10 in the container images.


ccoenen:
scrypt should also have failed unless this has been fixed in the meantime.

SISheogorath:
While I agree that it should have failed, somehow it didn't ._. https://travis-ci.org/hackmdio/codimd/jobs/457083337#L476-L511

PR #1066, opened 2018-11-20
Switch scrypt library to a successor

SISheogorath:
Since our previous scrypt library is unmaintained since 3 years, it's
time to look for an alternative.

A refactoring towards another password algorithm was worked on and this
is probably still the way to go. But for now the successor of our
previous library should already be enough.

https://www.npmjs.com/package/scrypt (old library)
https://github.com/ml1nk/node-scrypt (new library)

### DIFF
--- a/lib/models/user.js
+++ b/lib/models/user.js
@@ -1,7 +1,7 @@
 'use strict'
 // external modules
 var Sequelize = require('sequelize')
-var scrypt = require('scrypt')
+var scrypt = require('@mlink/scrypt')
 
 // core
 var logger = require('../logger')
--- a/package.json
+++ b/package.json
@@ -111,7 +111,7 @@
     "readline-sync": "^1.4.7",
     "request": "^2.88.0",
     "reveal.js": "~3.7.0",
-    "scrypt": "^6.0.3",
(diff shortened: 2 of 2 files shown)


ccoenen:
I do not believe there is a better algorithm at the moment than scrypt.
Also, the current storage scheme is the default for scrypt, so I wouldn't change this without looking at at length at a new implementation.

So in short:
- new (node-)scrypt: :tada: 
- Different algo: :thinking:

SISheogorath:
Well, I was looking towards argon2id as algorithm. At least put it into consideration.

We should implement some more flexibility into the password management either way, because if we want to upgrade the algorithm at any point in future, we should already have a way to do that. As this would also allow us to switch libraries without problems later on.

We'll see. I'm not sure if I want to work on the frontend code for the next release ([`1.4.0`](https://github.com/hackmdio/codimd/milestone/20)) or the authentication system.

ccoenen:
It should be noted that the ml1nk scrypt code differs from tarsnaps original slightly. Both changes seem to focus around MSVC, so it's probably Windows compatibility. Still, I'd love this to go upstream. This does not block us, but we should keep an eye on this.

PR #1067, opened 2018-11-20
Fix some linting warnings in `./lib/config`

SISheogorath:
We decided to switch to eslint for linting. This caused some new linter
warnings.

This patch fixes them for `./lib/config`


SISheogorath:
Should be done as a single PR.

PR #1068, opened 2018-11-20
Fix eslint warnings for database related libs

SISheogorath:
We introduced eslint as new linter to replace "`standard`". This causes
various warnings.

This patch fixes the warnings for `./lib/migrations` and `./lib/models`.


SISheogorath:
Should be done as a single PR.

PR #1069, opened 2018-11-21
Update from to-markdown to turndown

SISheogorath:
We got a security alert for a regular expression DoS attack on our used
library `to-markdown`.

After checking `to-markdown` to be maintained or not, it turned out they
renamed the library to `turndown`. So upgrading to `turndown` should fix
this vulnerbility.

References:
https://www.npmjs.com/package/to-markdown
https://github.com/domchristie/turndown/wiki/Migrating-from-to-markdown-to-Turndown


SISheogorath:
Moved to 1.4.0 due to missing testing in detail.

Since the function is only used for imports, the likelihood an attack with this regex DoS is close to zero. Because reloading the tab will solve the problem.

PR #1078, opened 2018-11-26
Add config for toobusy middleware

### LINKED ISSUE #1077: On slow~ish boxes, toobusy-js kicks in too early
Hey there,

I’m running CodiMD happily on a slightly aged server … happily? Well, almost! Sometimes, especially when some other services decide they need to trash the pool little DB hard, CodiMD reacts somewhat sluggish. In fact so sluggish that the default `maxLag` of `70ms` (?) of `toobusy-js` often kick in. This is slightly unnerving to users, especially on mobile where some browsers inhibit force-reload for some time after a failed load.

It would be great if `maxLag` would be configurable in some way for users like me. Of course, I guess we`ll die out eventually, replacing our old, steam-coughing servers with shiny new serverless-hybrid-hyperconverged-cloudnative-supercomputing instance-containers, but … until then, it’d be great if we could help ourselves via configuration instead of code-changes ^_^.

Best, and thanks for the great project!
-Dario

SISheogorath:
With very low CPU frequency or bad IO situation, as well as not-loaded
JS CodiMD happens to present unneeded "I'm busy"-messages to users.

This patch allows to configure the lag while keeping the libraries
defaults.

Fixes #1077

### DIFF
--- a/lib/config/default.js
+++ b/lib/config/default.js
@@ -56,6 +56,8 @@ module.exports = {
   // socket.io
   heartbeatInterval: 5000,
   heartbeatTimeout: 10000,
+  // too busy timeout
+  tooBusyLag: 70,
   // document
   documentMaxLength: 100000,
   // image upload setting, available options are imgur/s3/filesystem/azure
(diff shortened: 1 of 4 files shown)


lib/config/default.js
```diff
@@ -56,6 +56,8 @@ module.exports = {
   // socket.io
   heartbeatInterval: 5000,
   heartbeatTimeout: 10000,
+  // too busy timeout
+  tooBusyLag: undefined,
```
ccoenen: I think we should add 70 in this place.

* It's more self-documenting
* If the default was ever changed (no, I'm not assuming that it does), this will not get out of with our readme.

PR #1081, opened 2018-11-27
Fix typo

cloudyu:


ccoenen:
Hi! Would you like to elaborate on your changes? As far as I can see, one of the three commits is part of #1082 already. The "limited" change looks like it limits reading to the owner - which is not intended. Please refer to the [permissions table](https://demo.codimd.org/features#Permissions), only "private" is supposed to lock everyone else out.

cloudyu:
The latter two commits are meant to implement my own desired functionality. Because the wrong branch was submitted. Please ignore them

cloudyu:
The latter two commits are meant to implement my own desired functionality. because the wrong branch was submitted. Please ignore him

ccoenen:
No problem :-) Thanks for clarifying.

PR #1084, opened 2018-11-27
Prevent subdirectories in user export

### LINKED ISSUE #1079: [export user data] Slashes in note titles result in subdirectories
I have a note that is called `Philips Power2Go SCE4430/10`. On using the "export user data" feature, this file is put at this path in the archive:

```
/Philips Power2Go SCE4430/10.md
```

Is this the intended behaviour?

dsprenkels:
This commit also refactors the code a bit:

- It adds a `'-'` separator between a exported file's basename and its duplicate index. Example: `hello1.md` now becomes `hello-1.md`.
- It changes the implementation of an "already seen" set from using a list to using an object (because object indexing is O(1) instead of O(n).

This PR fixes #1079.


lib/web/userRouter.js
```diff
@@ -90,18 +90,20 @@ UserRouter.get('/me/export', function (req, res) {
           ownerId: user.id
         }
       }).then(function (notes) {
-        let list = []
+        let filenames = {}
```
ccoenen: good call renaming this :-)

lib/web/userRouter.js
```diff
@@ -90,18 +90,20 @@ UserRouter.get('/me/export', function (req, res) {
...
+          let filename
+          let suffix = ''
           do {
-            title = note.title + extension
-            extension++
-          } while (list.indexOf(title) !== -1)
+            let seperator = typeof suffix === 'number' ? '-' : ''
+            filename = basename + seperator + suffix + '.md'
+            suffix++
+          } while (filenames[filename] !== undefined)
```
ccoenen:
```suggestion
          } while (filenames[filename])
```

ccoenen: (this is really nitpicking, if it stays the same and we merge it, it's really fine as well)

dsprenkels: Updated the code. :)

PR #1095, opened 2018-12-11
Hide default ports from minio image upload urls

### LINKED ISSUE #1094: Remove port from Minio upload URL
I'm running Minio behing Traefik (with HTTPs enabled), therefore I specified `443` as `CMD_MINIO_PORT`.
When uploading an image
```md
![](https://myminio.backend:443/mybucket/uploads/upload_<id>.jpg)
```
is inserted into the editor.
Is there an option to remove the port here, so that
```
![](https://myminio.backend/mybucket/uploads/upload_<id>.jpg)
```
is inserted instead?

Thor77:
With the current state this behaviour is enabled by default and can be disabled by setting
```json
{
  "minio": {
    "hideDefaultPorts": false
  }
}
```

or `CMD_MINIO_HIDE_DEFAULT_PORTS=false`.

Fix #1094

### DIFF
--- a/lib/web/imageRouter/minio.js
+++ b/lib/web/imageRouter/minio.js
@@ -40,7 +40,7 @@ exports.uploadImage = function (imagePath, callback) {
         callback(new Error(err), null)
         return
       }
-      let hidePort = (protocol === 'http' && config.minio.port === 80) || (protocol === 'https' && config.minio.port === 443)
+      let hidePort = [80, 443].includes(config.minio.port)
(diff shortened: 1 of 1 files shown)


SISheogorath:
I think we can omit the config setting in favour of just removing the port when it's 443 or 80. I don't expect people to run HTTP on 443 and HTTPS on port 80.  If they do, we'll tell them that it's not supported.

Thor77:
Replaced the protocol + port check with one for port being `80` or `443`.

a60814billy:
Thanks @Thor77

PR #1096, opened 2018-12-12
Add plantuml support

chulkilee:
#362 

Note that it uses the public plantuml server.

<pre>
# PlantUML example

Following block will be rendered

@startuml
Alice -> Bob: Authentication Request
Bob --> Alice: Authentication Response

Alice -> Bob: Another authentication Request
Alice <-- Bob: Another authentication Response
@enduml

Following code fence block will not be rendered

```
@startuml
Alice -> Bob: Authentication Request
Bob --> Alice: Authentication Response

Alice -> Bob: Another authentication Request
Alice <-- Bob: Another authentication Response
@enduml
```
</pre>


ccoenen:
Nice work! The only thing that I am somewhat reluctant about is the syntax. I know this is largely due to `markdown-it-plantuml`, but it adds yet another markup element. Once we add something like this, we are pretty much stuck with it.

At the moment, extensions mostly use the code block syntax, like this:

    ```mermaid
       (some mermaid code)
    ```

How much effort would be needed to change it to use that same syntax?

chulkilee:
@ccoenen I actually prefer that too :heart:

<pre>
```plantuml
(plantuml code)
```
</pre>

This would work better with other markdown toolings, such as [vscode plugin](https://github.com/qjebbs/vscode-plantuml) and [PlantUML in GitLab](https://docs.gitlab.com/ee/administration/integration/plantuml.html)

I found the vscode plugin does it with "rule" and "render"

- https://github.com/qjebbs/vscode-plantuml/blob/master/src/markdown-it-plantuml/rule.ts
- https://github.com/qjebbs/vscode-plantuml/blob/master/src/markdown-it-plantuml/render.ts

It also has a code to generate code for URL at https://github.com/qjebbs/vscode-plantuml/blob/master/src/plantuml/renders/plantumlServer.ts

I asked for extracting the plantuml code in it into separate package (instead copying the code into here) - https://github.com/qjebbs/vscode-plantuml/issues/180 .. and found that there is [plantuml-encoder](https://www.npmjs.com/package/plantuml-encoder). So copied minimal code from the vscode plugin (which is in MIT license).

Probably it needs more work for other view (e.g. pdf, download), docs, and other type of plantuml, and different image format (svg vs png). But the basic feature is working.

public/js/extra.js
```diff
@@ -1041,6 +1043,33 @@ md.renderer.rules.fence = (tokens, idx, options, env, self) => {
...
+const makePlantumlURL = (umlCode) => {
+  let format = 'svg'
+  let code = plantumlEncoder.encode(umlCode)
+  return `${plantumlServer}/${format}/${code}`
+}
+
+// https://github.com/qjebbs/vscode-plantuml/tree/master/src/markdown-it-plantuml
+md.renderer.rules.plantuml = (tokens, idx) => {
+  let token = tokens[idx]
+  if (token.type !== "plantuml") {
```
jackycute:
```suggestion
  if (token.type !== 'plantuml') {
```

public/js/extra.js
```diff
@@ -1041,6 +1043,33 @@ md.renderer.rules.fence = (tokens, idx, options, env, self) => {
...
+    return tokens[idx].content
+  }
+
+  let url = makePlantumlURL(token.content)
+  return `<img src="${url}" />`
+}
+
+// https://github.com/qjebbs/vscode-plantuml/tree/master/src/markdown-it-plantuml
+md.core.ruler.push("plantuml", (state) => {
```
jackycute:
```suggestion
md.core.ruler.push('plantuml', (state) => {
```

public/js/extra.js
```diff
@@ -1041,6 +1043,33 @@ md.renderer.rules.fence = (tokens, idx, options, env, self) => {
...
+  let blockTokens = state.tokens
+  for (let blockToken of blockTokens) {
+    if (blockToken.type === "fence" && blockToken.info === "plantuml") {
```
jackycute:
```suggestion
    if (blockToken.type === 'fence' && blockToken.info === 'plantuml') {
```

public/js/extra.js
```diff
@@ -1041,6 +1043,33 @@ md.renderer.rules.fence = (tokens, idx, options, env, self) => {
...
+        blockToken.type = "plantuml";
```
jackycute:
```suggestion
      blockToken.type = 'plantuml'
```

jackycute:
Hi @chulkilee, thank you for your contribution!

There are some warning from the linter: https://travis-ci.com/hackmdio/codimd/jobs/221290183
Could you take a look of my suggestions?

chulkilee:
@jackycute thanks for the catch!

Pushed changes

- fix lint error
- rebase onto `develop`
- add a separate commit to fix `yarn.lock`

public/js/extra.js
```diff
@@ -1036,6 +1038,33 @@ md.renderer.rules.fence = (tokens, idx, options, env, self) => {
   return `<pre><code${self.renderAttrs(token)}>${highlighted}</code></pre>\n`
 }
 
+const makePlantumlURL = (umlCode) => {
+  let format = 'svg'
+  let code = plantumlEncoder.encode(umlCode)
+  return `${plantumlServer}/${format}/${code}`
```
Yukaii: Access explicitly from window can remind us where it's from, and the eslint global rule can also be removed after this changed.

```suggestion
  return `${window.plantumlServer}/${format}/${code}`
```

jackycute: @chulkilee could you consider this?

chulkilee: I don't know how that config is injected to windows. It seems like that (global, as the top eslint rule says).

However, as there is already `serverUrl` - I was confused - that why `plantumlServer` is special here.

Probably config should be exposed as `config` - and we should use `window.config.xxx`, instead of `window.xxx`. I believe that should be a separate PR.

Yukaii: Ok I see 👍, We can do this refactor later, providing cleaner config interface. Let's merge this PR @jackycute !

jackycute:
Thanks @chulkilee @Yukaii

public/js/extra.js
Yukaii: BTW, global accessible constants is loaded via script tag https://github.com/hackmdio/codimd/blob/6bce9ac5bf4d98f65f55621ea5073781328ba2a1/public/views/includes/scripts.ejs#L1

which generated in https://github.com/hackmdio/codimd/blob/1d7895c8693e30e4291576d293a9998db6d461c0/lib/web/statusRouter.js#L94-L110

blademainer:
Great job!

jcalfee:
Thanks for plantuml.  I'm really enjoying it.  The plantuml server locks me out every hour due to too many requests.  I was not aware this depended on the demo server.  The lockout is only for few minutes.  It may be due also to having several different umls in one document and a fast update time after I type something.  Increasing is just a patch, if several people are viewing or updating that is going to hit the demo server's quote sooner.

jackycute:
> Thanks for plantuml. I'm really enjoying it. The plantuml server locks me out every hour due to too many requests. I was not aware this depended on the demo server. The lockout is only for few minutes. It may be due also to having several different umls in one document and a fast update time after I type something. Increasing is just a patch, if several people are viewing or updating that is going to hit the demo server's quote sooner.

Hi @jcalfee, in this case, I would suggest self host a plantuml service and point `CMD_PLANTUML_SERVER` to the service.

pihentagy:
Is it possible to configure the plantuml server used? I'd like to use my own plantuml server.

jackycute:
> Is it possible to configure the plantuml server used? I'd like to use my own plantuml server.

Yes, please set env var `CMD_PLANTUML_SERVER` to your self hosted plantuml endpoint.

PR #1097, opened 2018-12-12
Added a configuration option 'disableRequestedAuthnContext' for passport-saml

nopap:
Added a configuration option 'disableRequestedAuthnContext' for passport-saml
disableRequestedAuthnContext: true|false

By default only Password authmethod is accepted, this option allows any other method.

Issue and option described here:
https://github.com/bergie/passport-saml/issues/226

Signed-off-by: Emmanuel Ormancey <emmanuel.ormancey@cern.ch>


jackycute:
Thanks for your contribution @nopap

PR #1108, opened 2018-12-21
Update upload provider error message

### LINKED ISSUE #1107: Bad error message in `lib/config/index.js`
https://github.com/hackmdio/codimd/blob/3cfd18d54f791b45c5d5e0114cdc18a5e941765d/lib/config/index.js#L166-L170

[^]: Line 168 mentions that the settings defaults to `'imgur'`, but in reality `'filesystem'` is being chosen. This should be fixed.

This bug was introduced in https://github.com/hackmdio/codimd/commit/972a81aa6fb2fb3883f7877c9c141cb68b50b871. I am happy to fix it myself (and update the warning message to `'filesystem'`).

dsprenkels:
972a81aa updated the actual default. This PR reflects that change in the preceding error message.

Fixes #1107.

### DIFF
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -165,7 +165,7 @@ if (config.sessionSecret === 'secret') {
 
 // Validate upload upload providers
 if (['filesystem', 's3', 'minio', 'imgur', 'azure'].indexOf(config.imageUploadType) === -1) {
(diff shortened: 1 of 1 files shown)


SISheogorath:
Thanks a lot!

PR #1112, opened 2018-12-28
Fix some XSS issues

jackycute:
- Fix to sanitize disqus shortnames to remove slashes [Security Issue]
- Fix to escape html comment tag [Security Issue]


SISheogorath:
Thanks a lot!

PR #1120, opened 2019-01-18
Drop node 6.x support

SISheogorath:
Seems like one of our deep dependencies dropped node version 6. This
results in a failed execution of yarn, that refuses to install the
library.

Since we already deprecated version 6 in earlier releases, dropping it
is a valid option.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,12 +10,6 @@ jobs:
   include:
     - env: task=npm-test
       node_js:
-        - 6
-      before_install:
-        - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version "$YARN_VERSION"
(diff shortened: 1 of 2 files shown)


ccoenen:
introduced by https://github.com/salesforce/tough-cookie/pull/138 / related to https://github.com/salesforce/tough-cookie/issues/140

ccoenen:
Voting to push this to 1.4.0.

Also: when merging this, this commit can also be reverted:
https://github.com/hackmdio/codimd/pull/1125/commits/fa0dea0a1b192ddff54fff5b46df930fd3d12195

ccoenen:
I rebased this onto current master and reverted that other node 6 compatibility fix (that's no longer needed).

Result is here: https://github.com/ccoenen/CodiMD/tree/remove/nodeV6.x

PR #1124, opened 2019-01-22
Fix reference to SAML guide in README

phrix32:


SISheogorath:
Good catch! Thanks a lot!

Issue #1128, opened 2019-01-24
Have S3 DOMAIN env setting so that Amazon URLs are not exposed


bmann:
I have Amazon S3 uploads setup in CodiMD and it all works great.

I have a bucket named `images.example.com`, and I even have DNS setup so that `images.example.com` is a CNAME for my Amazon bucket.

As far as I can tell, there is no S3_DOMAIN environment setting, so all the image URLs look like `https://s3-ca-central-1.amazonaws.com/images.example.builders/uploads/upload_d0e2a4d44acf975c9b748ce7d47f5a94.png`

(which is REGION/BUCKET)

The feature I am suggesting would introduce a new S3_DOMAIN env variable, where I could set `images.example.com` (ideally with / without https) and then the image URLs would be `images.example.com/uploads/upload_d0e2a4d44acf975c9b748ce7d47f5a94.png`.

If this feature DOES exist already, help in understanding how to set it would be appreciated :)

SISheogorath:
It doesn't exist right now, but we might get that in the next version when you want to spend some time on it.

The main part to modify is here:

https://github.com/hackmdio/codimd/blob/3dc40116e4ce8b6eeba6dc1c972dac74b518cb46/lib/web/imageRouter/s3.js#L44-L48

And adding the option to the config library. PRs are welcome :+1:

bmann:
Great, thank you! I will put it on my backlog to work on :) Thanks for the very direct pointer to the code!

jackycute:
Close in favor of #1142

PR #1130, opened 2019-01-25
Adds support for S3 Domains

bmann:
Adds a new config setting, S3_DOMAIN, that allows for images hosted on S3 to be hosted from a domain `images.example.com`. Only serves over https, so user will need to configure SSL using Cloudfront or Cloudflare.

imageRouter / s3, config, and README updates

Implements #1128


jackycute:
I think this duplicated with #1142

jackycute:
Thanks for your contribution @bmann

PR #1139, opened 2019-02-07
Corrected a typo

Luclu7:
Signed-off-by: Luclu7 <me@luclu7.fr>


SISheogorath:
Thanks for your contribution! Looking forward for more!

Usually we use prefer translations by using [POEditor](https://translate.codimd.org) as our translation platform. You would be very welcome to join us there!

PR #1150, opened 2019-02-21
Remove broken speakerdeck embedding

SISheogorath:
The current speakerdeck implementation is broken. An alternative
implementation using oembed doesn't work due to CORS, which could be
solved by proxying the speakerdeck API, but we decided to not do this.

This patch provides the link to the speakerdeck presentation instead,
and this way doesn't break existing notes. This is right now the best
solution we could come up with.


ccoenen:
Linking to the document is a good idea.

PR #1153, opened 2019-02-27
Fix empty serverURL did not redirect properly

toshi0123:
If serverurl is empty and session was expired, redirect doesn’t properly.

302 Found
Location:


SISheogorath:
Thanks for your work! Looking forward to more contributions! :smile:

PR #1158, opened 2019-03-04
Add serbian language

SISheogorath:
Thanks for the work of the translator Vladan we got a Serbian
translation added! Those few changes will add Serbian language support
for future CodiMD releases.


cvladan:
Thanks ;)

public/views/index/body.ejs
```diff
@@ -148,6 +148,7 @@
                         <option value="da">dansk</option>
                         <option value="ko">한국어</option>
                         <option value="id">Bahasa Indonesia</option>
+                        <option value="sr">српски</option>
```
SISheogorath: @cvladan can you confirm this name for the language is correct?

cvladan: I confirm it is proper.
But I do notice that not all languages have the same first letter casing (uppercase or lowercase).

For example, I'm quite sure "italiano" should be "Italiano", for Croatian, **"hrvatski jezik" should be only "Hrvatski"** as "jezik" means "language", **"język polski" should be only "Polski"**, "svenska" to "Svenska", "dansk" to "Dansk", and **in that light, "српски" should be "Српски".**

SISheogorath: Thanks a lot! Will be fixed once https://github.com/hackmdio/codimd/pull/1180 is merged, thanks!

PR #1161, opened 2019-03-05
Revert deletion of cookie "expires"

toshi0123:
** This is draft. **
I'm checking another part of express-session.
---

This change is workaround for express-session.
I read [an issue](https://github.com/expressjs/session/issues/365) about session cookies.
But I think there is some bugs about using only "maxAge".

### DIFF
--- a/app.js
+++ b/app.js
@@ -138,7 +138,8 @@ app.use(session({
   saveUninitialized: true, // always create session to ensure the origin
   rolling: true, // reset maxAge on every response
   cookie: {
-    maxAge: config.sessionLife
+    maxAge: config.sessionLife,
+    expires: new Date(Date.now() + config.sessionLife)
   },
   store: sessionStore
 }))


jackycute:
Hi @toshi0123 
Could you describe what's the bug using only `maxAge`?
If there is a bug beneath express-session, we should report to them.

toshi0123:
@jackycute Sorry, I need to clarify my problem.
I found expiring sessions frequently at my codimd. (about one hour or less)
So I checked express-session and modified the cookie part.
It seems worked, but it isn't.

I'll close this PL.

The thing is many new sessions (ex. handreds of new sessions) comes, then the valid session will expire.
I don't figure out the problem part right now.

jackycute:
Could you check the cookies in your browser?
It will show the expire time in the console.
Also, what kind of proxy are you using?
Or you might install any browser extension or script that might affect the cookies?

toshi0123:
I checked the cookie at developer tool in my chrome.
It shows about 2 weeks later.
I can write the codimd note normaly.
But I move to codimd top page, codimd shows me to signin window.
(In the background, the many new sessions comes.)

So I guess there is the problem in codimd top page.

SISheogorath:
@toshi0123 Do you see a warning about the session secret on startup? In this case your session secret is not set and this way a new random one is generated on each start. This will always cause a logout from all sessions of the previous CodiMD process.

https://github.com/hackmdio/codimd/blob/de0acbb566b3a9be095c414bf15edb734ac755be/lib/config/index.js#L158-L164

The solution is also rather simple. Either setting `CMD_SESSION_SECRET` to something long and random. Or using the `config.json` to configure it using `sessionSecret`.

toshi0123:
I setted the `CMD_SESSION_SECRET` already.
So I think the random sessionSecret is not generated.

I haven't been able to reproduce the problem at another environment.
So I updated the our production 1.2.1-alpine to 1.3.0-alpine.
It seems gone right now.

Thank you very much, all! I'll figure out the problem next time!

toshi0123:
I found my problem.
Our production database was corrupted.
The sequelize findOne function sometimes returned null data.
So I rebuilded our postgresql.

ccoenen:
Thanks for reporting back! I hope you didn't lose any data!

PR #1162, opened 2019-03-05
add feature: elapsed-time-bar in slides mode

pastleo:
### Changes

add https://github.com/tkrkt/reveal.js-elapsed-time-bar

Peek 2019-04-05 15-54

### Usage

at top of the note:

```
---
slideOptions:
  allottedMinutes: 5
---
```

then open slides mode, there should be a elapsed time bar


SISheogorath:
Hey @pastleo it's great to see that someone puts a lot of work into our slide mode! The changes you made are quite interesting, but it seems to be mainly focused on the slide mode. Since slide most is mostly a wrapper for [reveal.js](https://github.com/hakimel/reveal.js), I wonder if you tried to push your improvements there.

As while I think that those features seem quite good, I think they should be implemented upstream so all reveal.js projects can profit. Not completely sure you agree with me, but I hope it makes sense.

Also let's work a bit on the UX of the feature. Allowing those features only by key binding is something that is quite common for reveal, but I think we should try to do it better. Let's see if we can find a place to have a button for those features somewhere. (I think the speaker notes would be the right place, currently activated by pressing `s`). It also has a timer features but mainly to inform the speaker about how long they are already talking. Maybe setting a countdown timer by [slide options](https://demo.codimd.org/yaml-metadata#slideOptions) would be a way to go for a count down.

Again, I think the features are a great work, let's just make sure they make it into the right place and be available for everyone :)

ccoenen:
@pastleo sorry, I dropped the ball on this one. I was talking to @SISheogorath about this a few days ago and I promised to post an answer. I didn't. I'm sorry.

pastleo:
after reading [plugin guidelines](https://github.com/hakimel/reveal.js/wiki/Plugin-Guidelines) and [the plugin list](https://github.com/hakimel/reveal.js/wiki/Plugins,-Tools-and-Hardware) from revealjs wiki, I might reinvented the wheel: https://github.com/tkrkt/reveal.js-elapsed-time-bar

if I have time in the following days, I will try to integrate reveal.js-elapsed-time-bar instead

SISheogorath:
Awesome! Looking forward to your contribution :)

PR #1178, opened 2019-03-20
Updating french locale

ajanvier:
Just a little update of the french locale file : fixing some typos and adding two new translations.

(I hope I'm doing it the correct way, I apologize if it's not the case)


ajanvier:
already updated by POEditor.com, sorry for the inconvenience

PR #1179, opened 2019-03-24
Update history on note creation

cdancette:
When creating a new note with the API, the note doesn't appear in the history. 
If the URL returned by the API is lost, or never accessed, then the note is lost, and the owner can't access it.

This pull request updates the history on note creation.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -14,6 +14,7 @@ var config = require('./config')
 var logger = require('./logger')
 var models = require('./models')
 var utils = require('./utils')
+var history = require('./history')
 
 // public
 var response = {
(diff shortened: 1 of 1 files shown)


SISheogorath:
This looks overly familiar :) The old implementation pushed updates for the history from client side, and still exists: https://github.com/hackmdio/codimd/blob/50c80c99a4246c60026517b8632a1bbb05e1fe5c/lib/web/historyRouter.js#L9-L18

This is handled here: https://github.com/hackmdio/codimd/blob/50c80c99a4246c60026517b8632a1bbb05e1fe5c/lib/history.js#L138-L177

I think this should hit the same purpose.

a60814billy:
Thanks @cdancette

PR #1181, opened 2019-03-25
Move upload button into toolbar

SISheogorath:
Currently we have the odd situation to have two toolbars. One inside the
header and one in the editor.

Since we only show the image upload button when the editor is visible we
should move the upload button into the editor toolbar.

This patch does this by adding the image upload button besides the image
tag button.

### DIFF
--- a/public/js/index.js
+++ b/public/js/index.js
@@ -304,7 +304,6 @@ var editor = editorInstance.init(textit)
 // FIXME: global referncing in jquery-textcomplete patch
 window.editor = editor
 
-var inlineAttach = inlineAttachment.editors.codemirror4.attach(editor)
 defaultTextHeight = parseInt($('.CodeMirror').css('line-height'))
 
 //  initalize ui reference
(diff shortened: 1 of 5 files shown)


SISheogorath:
If you wonder how the changed version looks like:

image

ccoenen:
absolutely. There's other things that need to switch places, too, but this is as good a place to start as any.

SISheogorath:
I recently thought about how to streamline the UI a bit. the night mode should move into the user profile settings, as soon as they are done. (until then, it'll stay where it is, there is no point in moving things twice).

The "help" should move into the drop down menu. Same goes for the "publish"-button.

Currently not sure where to put the node permissions, but they'll also stay in place until we rework the permission system (or someone has a great idea).

ccoenen:
not sure about the night mode. I'm not using it in CodiMD at all, but on my phone I use similar functionality quite often, and there I switch around whenever I feel like it. I'd be interested how other people use this.

I have some ideas for a general re-ordering of ui elements as well. I'll write them up at some point.

Issue #1187, opened 2019-04-08
js-sequence-diagrams not found


LuckyC4t:
First I run `git clone https://github.com/hackmdio/codimd.git`
then i run`cd codimd && bin/setup`, it show errors
```
copy config files
install packages
yarn install v1.15.2
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
error An unexpected error occurred: "https://registry.yarnpkg.com/js-sequence-diagrams/-/js-sequence-diagrams-1000000.0.6.tgz: Request failed \"404 Not Found\"".
info If you think this is a bug, please open a bug report with the information provided in "/codimd/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```
even if i ignore it and run `npm run build`, this error still exists.

`node v8.10.0, npm 3.5.2, yarn 1.15.2`

a60814billy:
js-sequence-diagram was removed by npm several days ago, we will fix it as
soon as possible.

LuckyCat <no-reply@hackmd.io>於 2019年4月8日 週一，10:36寫道：

a60814billy:
Hi @LuckyC4t, this issue already fixed on master branch, please try to pull newest code and test it.
let me know if you have any problem.

Luzifer:
Any chance to get a release with the fix soon? Sadly the current release 1.3.1 is still affected and broken.

PR #1188, opened 2019-04-10
Fix/js sequence diagram

a60814billy:
Fixed: #1187 

## Todo: 
- [x] fix js-sequence-diagram
- [x] regression error (flowchart.js)
- [x] js-sequence-diagram too height


jackycute:
Tested and LGTM!

PR #1193, opened 2019-04-14
Fix stored XSS in the graphviz error message rendering [Security Issue]

jackycute:
There is vulnerability inside view rendering that might produce XSS.

Thanks @k1tten for reporting to us 😄


k1tten:
Also thanks to @HexRabbit  🎉

PR #1211, opened 2019-05-03
Adding an error code to the `{"status":"forbidden"}` response

gndu91:
I think a 401 status code should be more appropriate than 200 when the user is not recognized.


a60814billy:
thanks @gndu91

PR #1228, opened 2019-06-27
Refactor realtime

a60814billy:


lib/realtimeClientConnection.js
```diff
@@ -0,0 +1,232 @@
...
+  async changeNotePermission (newPermission) {
+    const changedRows = await models.Note.update({
+      permission: newPermission
+    }, {
+      where: {
+        id: this.getCurrentNote().id
+      }
+    })
+    if (changedRows !== 1) {
+      throw new Error(`updated permission failed, cannot set permission ${newPermission} to note ${this.getCurrentNote().id}`)
```
jackycute: I tried to change permission of the note, but see this error in the log.

lib/realtimeClientConnection.js
```diff
@@ -0,0 +1,232 @@
...
+
+  async changeNotePermission (newPermission) {
+    const changedRows = await models.Note.update({
+      permission: newPermission
+    }, {
+      where: {
+        id: this.getCurrentNote().id
+      }
+    })
+    if (changedRows !== 1) {
```
jackycute: This should be

```suggestion
    if (changedRows[0] !== 1) {
```

jackycute: This might need a test case.

jackycute:
LGMT, thanks @a60814billy

PR #1152, opened 2019-02-23
Add allowAnonymousViews option

hakoai:
resolve #1144
allowAnonymousViews : set to `false` to delete `freely`, `editable` and `locked` permission when allowAnonymous is true (default is `true`)

e.g.
`allowAnonymous=false`
a-false

`allowAnonymous=false` and `allowAnonymousViews=false`
a-false-v-false

`allowAnonymous=false` and `allowAnonymousEdits=true`
a-false-e-true

| allowAnonymous | allowAnonymousEdits | allowAnonymousViews | `New guest note` button | `freely` permission | `editable/locked` permission | 
| -------- | -------- | -------- | -------- | -------- | -------- |
| true | - | - | active | active | active |
| false | - | false | inactive | inactive | inactive |
| false | false | true | inactive | inactive | active |
| false | true | true | inactive | active | active |

PR #1232, opened 2019-07-01
Block note creation by anonymous users if allowAnonymousEdits is false

blha303:
If anonymous users can't edit notes, they also shouldn't be able to create notes (in my opinion! I don't want to run a public pastebin). This patch should restrict guests from seeing the "New guest note" button on the index page, and sends them to a forbidden page if they try /new, on instances where allowAnonymous=true and allowAnonymousEdits=false.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -124,7 +124,7 @@ function newNote (req, res, next) {
   body = body.replace(/[\r]/g, '')
   if (req.isAuthenticated()) {
     owner = req.user.id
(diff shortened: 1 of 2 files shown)


jackycute:
Hi @blha303,
Thank you for your contribution!

I think you might want to take a look at PR #1152? It looks proposing more complete solution for this purpose.

jackycute:
Close in favor of #1152

PR #1239, opened 2019-07-18
Make toobusy.maxLag configurable

notake:
Make [toobusy-js](https://github.com/STRML/node-toobusy) maxLag parameter configurable with environment variables.
Default value is 70ms.


jackycute:
Thank you for your contribution! @notake

PR #1245, opened 2019-08-01
feat: change guest random nameing to chance.animal()

### LINKED ISSUE #1244: Stop assigning random names to guests
Currently guests are assigned a random surname:

https://github.com/hackmdio/codimd/blob/ba0e4df6b7284739610ac7f2c0d075c23c112c5c/lib/realtime.js#L583

This is cute but also confusing to users:

- When a team member with a real name is viewing a note as guest, they are given a new, fake name.
- It's easy to get confused and think that this is a real name - maybe this is a new team member I haven't met yet, or even an intruder?

I would suggest that Guests are given a different unique identifier. I'd suggest a truncated UUID (`Guest abc123`) or [`chance.animal`](https://chancejs.com/thing/animal.html).

a60814billy:
assigning random animal name to guest
resolved #1244

### DIFF
--- a/lib/realtime.js
+++ b/lib/realtime.js
@@ -580,7 +580,7 @@ function updateUserData (socket, user) {
     user.login = true
   } else {
     user.userid = null
-    user.name = 'Guest ' + chance.last()
+    user.name = 'Guest ' + chance.animal()
     user.login = false
   }
 }


jackycute:
Thanks @a60814billy

PR #1252, opened 2019-08-05
 fix: history api failed cause by circular dependency

a60814billy:
fix circular dependencies cause by PR #1179


jackycute:
Thanks @a60814billy

PR #1253, opened 2019-08-06
fix(security): potential JSONP callback overwritten

jackycute:
Fix potential JSONP callback overwritten in Slideshare/Vimeo/Gist external embed syntax.


jackycute:
Thanks @Yukaii

PR #1259, opened 2019-08-18
Remove BUILD_ASSETS environment variable

Yukaii:
### Changes

Remove unnecessary `BUILD_ASSETS` environment variable. We can check whether it runs on heroku by the `DYNO` env


jackycute:
Thanks @Yukaii

Issue #1263, opened 2019-08-21
Open Redirect / XSS via iframe with sandbox.


Ry0taK:
## Summary
This is similar issue to #959, but different cause.
Due to `sandbox` attribute is allowed, attacker can redirect victim to something malicious if attacker embeds iframe with `allow-top-navigation` in `sandbox` attribute. 
#959 won't work in Chrome because of their security, but this will work due to `sandbox` is whitelist attribute.
This will be XSS in Safari by using data scheme.
## Step to reproduce
1. Type ```<iframe src="https://exec.ga/hackmd.html" sandbox="allow-scripts allow-top-navigation allow-scripts"></iframe>``` in Editor.
2. View edited page.

[PoC](https://demo.codimd.org/YcIX0oKzQAuZMMOqxeASdg)
(This PoC will redirect you to [example.com](https://example.com))

## Suggested fix
Don't allow `sandbox` attribute in public/js/render.js.

Ry0taK:
Can I request CVE number for this vulnerability?

jackycute:
Hi @Ry0taK,
Thanks for reporting this to us.
Yes, I think you can report to CVE, this should be fixed soon.

Ry0taK:
Hello @jackycute 
Thank you for your reply.
I've requested CVE number to MITRE Corporation.

Regards,
RyotaK

SpyroTEQ:
Hi,
I'm not a user of this codebase project (I ended up here through the CVE RSS), but I don't get why removing/disallowing the sandbox attribute will protect nor change anything?

This, served at `http://victim.localhost/`
```html
<iframe src="http://attacker.localhost/contained.html">
</iframe>
```

With this evil contained page
```html
<script>
window.top.location.href = 'https://example.com/';
</script>
```

Will also redirect top page to `https://example.com/` ?! Why are things safer after the patch?

jackycute:
Hi @SpyroTEQ,

So disallowing sandbox attribute is the first step to prevent this vulnerability.
There is another issue to track #959

PR #1264, opened 2019-08-21
Remove sandbox attribute

Ry0taK:
Patch for #1263
Because sandbox is whitelist attribute, attacker will be able to create iframe that has more permission than default.
I'm not sure `sandbox` is commonly used, but at least this should be removed because of this leads XSS in some browsers.

### DIFF
--- a/public/js/render.js
+++ b/public/js/render.js
@@ -20,7 +20,7 @@ whiteList['style'] = []
 // allow kbd tag
 whiteList['kbd'] = []
 // allow ifram tag with some safe attributes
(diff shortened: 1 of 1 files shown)


jackycute:
LGTM, thanks @Ry0taK 
I think we need another PR for adding sandbox attribute to all iframes in the markdown HTML.

Ry0taK:
Hello @jackycute and thanks!

Should some value (e.g. allow-scripts) be set for the sandbox attribute, or empty value?
If you add `sandbox` attribute without any value, it won't allow script,forms... etc, so I think at least allowing script is better.

Regards,
RyotaK

jackycute:
I agree, we need to prevent iframe accessing top frame but should allowing scripts in the iframe.

PR #1301, opened 2019-10-15
Update develop

LeePorte:
We were a long way behind, this brings us up to date with upstream


LeePorte:
Apologies. opened against upstream rather than our fork

Issue #1302, opened 2019-10-17
Code syntax highlight on pdf is different from that on preview.


simkaren:
Hi,

when I publish my markdown document with source code as pdf, I found that the syntax highlight on pdf is different from that of on preview.
  
preview on browser
<img width="347" alt="image2" src="https://user-images.githubusercontent.com/55137477/66985324-862a0580-f0f7-11e9-9ad1-bfa1cbfcb5a7.png">
on pdf
<img width="751" alt="image1" src="https://user-images.githubusercontent.com/55137477/66985327-8924f600-f0f7-11e9-8505-5c254e751c11.png">

I think the syntax highlight on preview is better than that on pdf, so I want to change it.

Thank you.

simkaren:
The pull request is merged so I close this issue. Thank you.

PR #1308, opened 2019-10-21
Feature/syntax hilight on PDF

kondounagi:
related to #1302 

## Description
Set [markdown-pdf](https://www.npmjs.com/package/markdown-pdf)'s option to provide syntax highlight on PDF output.

I worked with @betsu6419 and @simkaren .
## Demo
Realize syntax highlight on PDF same to View.
### Before
image
### After
image

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -316,17 +316,22 @@ function actionPDF (req, res, note) {
   var content = extracted.markdown
   var title = models.Note.decodeTitle(note.title)
 
+  var highlightCssPath = path.join(config.appRootPath, '/node_modules/highlight.js/styles/github-gist.css')
+
   if (!fs.existsSync(config.tmpPath)) {
     fs.mkdirSync(config.tmpPath)
   }
-  var path = config.tmpPath + '/' + Date.now() + '.pdf'
+  var pdfPath = config.tmpPath + '/' + Date.now() + '.pdf'
   content = content.replace(/\]\(\//g, '](' + url + '/')
-  markdownpdf().from.string(content).to(path, function () {
-    if (!fs.existsSync(path)) {
-      logger.error('PDF seems to not be generated as expected. File doesn\'t exist: ' + path)
+  var markdownpdfOptions = {
+    highlightCssPath: highlightCssPath
+  }
+  markdownpdf(markdownpdfOptions).from.string(content).to(pdfPath, function () {
+    if (!fs.existsSync(pdfPath)) {
+      logger.error('PDF seems to not be generated as expected. File doesn\'t exist: ' + pdfPath)
       return errorInternalError(res)
     }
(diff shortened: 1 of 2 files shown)


kondounagi:
Pull rquest build was failed but I don't know why...

Yukaii:
Hi @kondounagi, Thanks for your contribution!

It seems like there's problem in our CI service, so I triggered a re-run. Anyway the tests are passed now. 😅 

There's still one thing that you need to change before we review this PR: Rebase your commit onto the `develop` branch, saw a few commits that's no related this PR

Screen Shot 2019-10-21 at 5 27 12 PM

Yukaii:
I've also created a heroku review app for this PR, you can play with it and make sure your changes don't break the existing features 👍 

https://codimd-feature-syntax-h-pjsiu7.herokuapp.com/

kondounagi:
#### CI
> It seems like there's problem in our CI service, so I triggered a re-run. Anyway the tests are passed now. 

> I've also created a heroku review app for this PR, you can play with it and make sure your changes don't break the existing features

Thank you for your quick response!
Anyway, I was relieved to see this :relieved: 

#### Rebase
> There's still one thing that you need to change before we review this PR: Rebase your commit onto the develop branch, saw a few commits that's not related this PR

As you indicated, I rebased `kondounagi:feature/syntax-hilight-on-pdf` to `hackmdio:develop` :+1: 
image

lib/response.js
```diff
@@ -316,17 +316,23 @@ function actionPDF (req, res, note) {
   var content = extracted.markdown
   var title = models.Note.decodeTitle(note.title)
 
+  var appDir = path.dirname(require.main.filename)
+  var highlightCssPath = appDir + '/lib/hljs.css'
```
Yukaii: We put css assets under `public/css` directory as convention. Could you add one config above:

https://github.com/hackmdio/codimd/blob/56cc64e9109f1160d66a805ec693f07751fc5c1a/lib/config/index.js#L203

```js
config.appRootPath = appRootPath
```

So we can find `highlightCssPath` here by:

```js
var highlightCssPath = path.join(config.appRootPath, 'public/css/hljs.css')
```

Yukaii: Seems like `hljs.css` is taken from `/node_modules/highlight.js/styles/default.css`, we can use:

```js
var highlightCssPath = path.join(config.appRootPath, '/node_modules/highlight.js/styles/default.css')
```

(however the minification is lost 😅 )

If we make highlight.js theme configurable in the future, this can be done easier.

kondounagi: I fixed them as you pointed out :+1: 
(I set not `default.css` but `github-gist.css` to `highlightCssPath` because View page uses this theme!)

Yukaii:
Thanks for your contribution @kondounagi 😍

PR #1310, opened 2019-10-21
Allow to generate lower case header references through the config

### LINKED ISSUE #1305: Git(-Hub/Lab) compatible (lower-case) header references
The automatically generated header-references in Codimd are not compatible with github and gitlab (and others, like pandoc), because codi keeps the caseing of the text, while the standard is to make everything lower-case.

This is an issue if one transfers documents back and forth between codi and other systems, as internal links break.

Would it be possible to introduce a config variable to enable github compatible reference generation? or maybe even choosing completely different markdown flavors?

hoijui:
resolves #1305 

This makes the references consistent/compatible with GitHub,
GitLab, Pandoc and many other tools.
    
This behavior can be enabled in config.json with:
    
```
"linkifyHeaderStyle": "gfm"
```

### DIFF
--- a/config.json.example
+++ b/config.json.example
@@ -127,6 +127,7 @@
         "plantuml":
         {
           "server": "https://www.plantuml.com/plantuml"
-        }
+        },
+        "linkifyHeaderStyle": "gfm"
     }
 }
--- a/config.json.example
+++ b/config.json.example
@@ -3,7 +3,8 @@
         "db": {
             "dialect": "sqlite",
             "storage": ":memory:"
-        }
+        },
+        "linkifyHeaderStyle": "gfm"
     },
     "development": {
         "loglevel": "debug",
@@ -13,7 +14,8 @@
         "db": {
             "dialect": "sqlite",
             "storage": "./db.codimd.sqlite"
-        }
+        },
+        "linkifyHeaderStyle": "gfm"
     },
     "production": {
         "domain": "localhost",
--- a/lib/config/default.js
+++ b/lib/config/default.js
@@ -162,5 +162,6 @@ module.exports = {
   allowEmailRegister: true,
   allowGravatar: true,
   allowPDFExport: true,
-  openID: false
+  openID: false,
+  linkifyHeaderStyle: 'keep-case'
 }
--- a/lib/config/default.js
+++ b/lib/config/default.js
@@ -163,5 +163,18 @@ module.exports = {
   allowGravatar: true,
   allowPDFExport: true,
   openID: false,
+  // linkifyHeaderStyle - How is a header text converted into a link id.
+  // Header Example: "3.1. Good Morning my Friend! - Do you have 5$?"
+  // * 'keep-case' is the legacy CodiMD value.
+  //    Generated id: "31-Good-Morning-my-Friend---Do-you-have-5"
+  // * 'lower-case' is the same like legacy (see above), but converted to lower-case.
+  //    Generated id: "#31-good-morning-my-friend---do-you-have-5"
+  // * 'gfm' _GitHub-Flavored Markdown_ style as described here:
+  //    https://gist.github.com/asabaylus/3071099#gistcomment-1593627
+  //    It works like 'lower-case', but making sure the ID is unique.
+  //    This is What GitHub, GitLab and (hopefully) most other tools use.
+  //    Generated id:   "31-good-morning-my-friend---do-you-have-5"
+  //    2nd appearance: "31-good-morning-my-friend---do-you-have-5-1"
+  //    3rd appearance: "31-good-morning-my-friend---do-you-have-5-2"
   linkifyHeaderStyle: 'keep-case'
 }
--- a/public/js/extra.js
+++ b/public/js/extra.js
@@ -867,7 +867,11 @@ const linkifyAnchors = (level, containingElement) => {
     if (header.getElementsByClassName('anchor').length === 0) {
       if (typeof header.id === 'undefined' || header.id === '') {
         // to escape characters not allow in css and humanize
-        const id = slugifyWithUTF8(getHeaderContent(header))
+        var id = slugifyWithUTF8(getHeaderContent(header))
+        // to make compatible with GitHub, GitLab, Pandoc and many more
+        if (window.linkifyHeaderStyle !== 'keep-case') {
+          id = id.toLowerCase()
+        }
         header.id = id
       }
       if (!(typeof header.id === 'undefined' || header.id === '')) {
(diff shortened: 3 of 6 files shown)


public/js/extra.js
```diff
@@ -867,7 +867,11 @@ const linkifyAnchors = (level, containingElement) => {
     if (header.getElementsByClassName('anchor').length === 0) {
       if (typeof header.id === 'undefined' || header.id === '') {
         // to escape characters not allow in css and humanize
-        const id = slugifyWithUTF8(getHeaderContent(header))
+        var id = slugifyWithUTF8(getHeaderContent(header))
```
Yukaii:
```suggestion
        let id = slugifyWithUTF8(getHeaderContent(header))
```

public/js/extra.js
```diff
@@ -867,7 +867,11 @@ const linkifyAnchors = (level, containingElement) => {
...
+        // to make compatible with GitHub, GitLab, Pandoc and many more
+        if (window.linkifyHeaderStyle !== 'keep-case') {
```
Yukaii:
```suggestion
        if (window.linkifyHeaderStyle === 'gfm') {
```

Or any `linkifyHeaderStyle` value except `keep-case` would go into here (maybe that's your intent?)

hoijui: yes that was my intend, because everywhere else it is lower-case. the only reason to not move to lower-case as default - in my eyes - is backwards compatibility.

hoijui: I imagine people using any kind of Markdown flavor as value in this variable; examples from Pandoc:

* _markdown_ (Pandoc's Markdown)
* _markdown\_github_ (or _gfm_)
* _markdown\_mmd_
* _markdown\_phpextra_
* _markdown\_strict_

I was also already thinking about the possibility to one day choose the Markdown flavor for everything (not just the reference style). In that case, it makes sense to already have this feature designed in a somewhat compatible fashion, not requiring yet more backwards-compatibility precautions, for example.

public/js/lib/common/constant.ejs
```diff
@@ -6,4 +6,6 @@ window.plantumlServer = '<%- plantumlServer %>'
 
 window.allowedUploadMimeTypes = <%- JSON.stringify(allowedUploadMimeTypes) %>
 
+window.linkifyHeaderStyle = <%- JSON.stringify(linkifyHeaderStyle) %>
```
Yukaii: Here you can use:

```suggestion
window.linkifyHeaderStyle = '<%- linkifyHeaderStyle %>'
```

Yukaii:
Notice in the latest commit our DCO check is not passed. Please also update this commit https://github.com/hackmdio/codimd/commit/fdb81ee313ede2ce75e754c8930679de5d48c754 👍

hoijui:
thanks @Yukaii , I would not have noticed ;-)

config.json.example
```diff
@@ -127,6 +127,7 @@
         "plantuml":
         {
           "server": "https://www.plantuml.com/plantuml"
-        }
+        },
+        "linkifyHeaderStyle": "gfm"
```
jackycute: Hi @hoijui 
Could we put the available options of this settings in the comment?
I think people will edit this file for their first setup config.
Like:

```suggestion
        "linkifyHeaderStyle": "keep-case" // "keep-case" or "gfm"
```

hoijui: I also didn't know, but here are [no comments in JSON](https://stackoverflow.com/questions/244777/can-comments-be-used-in-json) :/

jackycute: Oh, you're right!

maybe this, but I don't know if people can understand this:
```suggestion
        "//linkifyHeaderStyle": "keep-case or gfm"
```

lib/config/default.js
```diff
@@ -162,5 +162,6 @@ module.exports = {
   allowEmailRegister: true,
   allowGravatar: true,
   allowPDFExport: true,
-  openID: false
+  openID: false,
+  linkifyHeaderStyle: 'keep-case'
```
jackycute: Hi @hoijui 
Could you add a comment in here to describe the possible options?
We will also add it into the documentation.

config.json.example
jackycute: I think maybe we can just remove this in the config.json.example.
It's better to add to the documentation: https://hackmd.io/c/codimd-documentation/%2Fs%2Fcodimd-configuration
We can help to do so.

jackycute:
Hi @hoijui,
Could you also add this config to https://github.com/hackmdio/codimd/blob/develop/lib/config/environment.js?
So that people can define this via environment vars.

Thanks!

lib/config/default.js
hoijui: yes!

config.json.example
hoijui: The idea was, to have it in there so it would be used by new installations, but not by old ones.
This way, we could make all new installations adhere to the standard, as used everywhere else.

jackycute: got you

hoijui:
ok, am trying.. how can I log any debug info?

Yukaii:
@hoijui You can insert `console.log` or add breakpoint in DevTools manually. Just remember do not commit those lines in VCS. :smile:

hoijui:
thanks! :-)
found that out late evening yesterday aswell.. got it almost working now, except: I need a list of all headers in the document, in order. I currently fetch them with `const allHeaders = view.find(`:header`).toArray()`, which does not give them in order of appearance.
How should I fetch them?

Yukaii:
You can use `sortByDepth` method like we used in building checkboxes

https://github.com/hackmdio/codimd/blob/fd155fa74dc613d22473b5f32ef94b6c3644f707/public/js/extra.js#L261

hoijui:
hmm.. it gives me the same order, with and without that sort.
I do not need them sorted by header level, but by the appearance of the header in the document.

sample document:

```mardown
# abc

### abc

## abc
```

I want them in the order they appear here, but both:

* `const allHeaders = view.find(`:header`).toArray()`
* `const allHeaders = view.find(`:header`).sortByDepth().toArray()`

returns them in this order:

```
# abc
## abc
### abc
```
which I guess, is sorted by (header-level/)depth.

Yukaii:
Screen Shot 2019-10-28 at 11 48 56 PM

@hoijui I couldn't reproduce in DevTool console, could you share the whole code snippet with me? Something like where did you insert this line.

hoijui:
thank you @Yukaii !
indeed, it was correct.. something else in my code went wrong.
I made 3 more commits, and it seems to work fine now, as far as my testing goes.

hoijui:
> Hi @hoijui,
> Could you also add this config to https://github.com/hackmdio/codimd/blob/develop/lib/config/environment.js?
> So that people can define this via environment vars.
> 
> Thanks!

did that too now

lib/config/default.js
hoijui: done here:
https://github.com/hackmdio/codimd/pull/1310/commits/3d5d8e20f73bf8d363975273f8ade70d729aa126

hoijui:
thanks you two! @Yukaii @jackycute 
you are really great at this! :-)

PR #1313, opened 2019-10-22
fix: image upload caused permission denied error in docker

a60814billy:
fixed wrong file permission setting in 1.4.0-rc.1 caused user can't upload image


jackycute:
Thanks @a60814billy

PR #1327, opened 2019-10-25
support to login with github enterprise

### LINKED ISSUE #1269: I want to use GitHub Enterprise with GitHub auth
We use hackmd for internal collaboration. I want to authenticate my account with github enterprise.

https://github.com/jaredhanson/passport-github#enterprise-corporate-github

kamijin-fanta:
resolve #1269

config example
```json
{
  "production": {
    "github": {
      "clientID": "**************",
      "clientSecret": "************************************",
      "enterpriseURL": "https://github-internal.example.com"
    }
  }
}
```

### DIFF
--- a/config.json.example
+++ b/config.json.example
@@ -50,7 +50,6 @@
             "consumerSecret": "change this"
         },
         "github": {
-            "enterpriseURL": "if you use github.com, not need to specify",
             "clientID": "change this",
             "clientSecret": "change this"
         },
--- a/config.json.example
+++ b/config.json.example
@@ -50,6 +50,7 @@
             "consumerSecret": "change this"
         },
         "github": {
+            "enterpriseURL": "if you use github.com, not need to specify",
             "clientID": "change this",
             "clientSecret": "change this"
         },
--- a/lib/config/environment.js
+++ b/lib/config/environment.js
@@ -66,6 +66,7 @@ module.exports = {
     consumerSecret: process.env.CMD_TWITTER_CONSUMERSECRET
   },
   github: {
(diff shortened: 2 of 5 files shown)


Yukaii:
Hi @kamijin-fanta , thanks for your contribution!

Could you please also update these two files as well?

- `lib/config/default.js`
- `lib/config/environment.js`

kamijin-fanta:
@Yukaii thank you for reviewing! it fixed now.

lib/config/environment.js
```diff
@@ -66,6 +66,7 @@ module.exports = {
     consumerSecret: process.env.CMD_TWITTER_CONSUMERSECRET
   },
   github: {
+    enterpriseURL: process.env.ENTERPRISE_URL,
```
Yukaii: We use `CMD` as environment variable prefix, so this should be `CMD_GITHUB_ENTERPRISE_URL`

kamijin-fanta: oops.... I was wrong. fixed.

config.json.example
```diff
@@ -50,6 +50,7 @@
             "consumerSecret": "change this"
         },
         "github": {
+            "enterpriseURL": "if you use github.com, not need to specify",
```
Yukaii: I think this line should be removed because we expect user copy this file as `config.js`. User will have to remove this line in config manually.

The comment part could be added in `lib/config/default.js` instead. 

@jackycute Just a reminder, we’ll also have to update our online documentation accordingly.

kamijin-fanta: thanks! it's fixed.

Yukaii:
Thank you @kamijin-fanta !

PR #1355, opened 2019-11-24
Multilanguage spellchecker

### LINKED ISSUE #647: Change spellcheck language
Is it possible to change the language of the spellchecker from english to german?

Yukaii:
resolves #647 

multilangauge-spellchecker

### TODOS

- [x] Load dictionary by language dynamically
- [x] remove num_loaded workaround
- [x] Bundle dictionaries to use locally
- [x] Load configured language when spellchecker initialized
- [x] organize code in `lib/editor/index.js`


Yukaii:
The current implementation works out well(see the gif), but there's a short freeze after the dictionary gets parsed(especially when the dictionary file is huge, e.g., German dict file is 1 MB while English is only at half). Doing some experiments in the commit 5d596fbb and it does the follows:

- Write a mini tokenizer for tokenizing each line
- Feed each line into tokenizer and break them into words
- Spellcheck each word with typo.js and map them into CodeMirror linter compatible formats (WIP)
- Do all the things in web worker thread with workerize-loader, for not locking the entire JavaScript UI thread (WIP).

Why choosing CodeMirror linter over exiting CodeMirror overlay? Just because it supports Promises and overlay does not. 

But the other problem is our markdownlint support already using linter plugin, and the CodeMirror builtin linter plugin needs to be extended to support **multiple** linting plugins. And this requires a larger rewrite of code.

### TLDR

I won't optimize this now. 😂

banderlog:
Should it work on https://hackmd.io/?

Yukaii:
Hi @banderlog, we're still evaluating this. Due to the fact that most of the dictionaries from https://github.com/wooorm/dictionaries are licensed under GPL, it cannot be used directly on hackmd.io. Also, which language do you want us to support first?

banderlog:
@Yukaii understood :)

russian

ad-m:
I am would like use Polish language. According to mentioned link Polish dictionary is licensed as:

> This dictionary for spell-checking Polish texts is licensed under
GPL, LGPL, MPL (Mozilla Public License) and Creative Commons
ShareAlike licenses (see http://creativecommons.org/licenses/sa/1.0).

PR #1371, opened 2019-12-05
Explicitely set uploaded image ACLs for S3 to "public-read"

djmaze:
This works around permission problems with DigitalOcean spaces (where files are always *private* by default).

This should not pose a difference with AWS and other providers as CodiMD only works with public S3 assets either way.


jackycute:
Thanks @djmaze 👍

PR #1261, opened 2019-08-19
Experimental support for customized emojify version

Yukaii:
related to #262 

### Description

Use [emoji-data](https://github.com/iamcal/emoji-data) to provide emoji sets (apple-flavored). It's more completed and standard compliance.

### Demo

Screen Shot 2019-09-08 at 12 29 08 AM

PR #1400, opened 2020-02-06
Support emojify.css with CDN

Yukaii:
In #1261, we switched emojify.js to our forked version but missed some works when enabling `useCDN` option. This PR finishes the rest of the work.


jackycute:
LGTM, Thanks @Yukaii

Issue #1402, opened 2020-02-06
Question - how to limit access the published notes to only signed-in users?


zulhfreelancer:
I have published a note to my self-hosted CodiMD instance and the shareable link looks like https://example.com/s/4OtuyRvuc

When I viewed the link in another browser (I'm not logged in as CodiMD user), I can see the note. I was expecting an error message saying I don't have view permission for that note.

I tried changing the note permissions (on owner side) but none of them worked. I'm still able to see the published notes on non-logged in browser.

How to only whitelist signed-in users to see the published notes? Did I missed something?

zulhfreelancer:
@jackycute appreciate if you can spare a minute to answer this :)

jackycute:
Hi @zulhfreelancer,

You need to change the note permission to either `Limited`, `Protected`.
For the permission details, please see: https://codimd-app.herokuapp.com/features#Permissions

zulhfreelancer:
@jackycute I already tried it. The published note is still visible to not-logged in users.

jackycute:
Can you reproduce that on https://codimd-app.herokuapp.com?
I tried this: https://codimd-app.herokuapp.com/hOoY0WwIR62WpcIyoO44-g
And I use incognito mode to visit that, got proper error message.

<img width="1028" alt="螢幕快照 2020-02-07 上午10 05 16" src="https://user-images.githubusercontent.com/2639151/73994741-93fae380-4991-11ea-821c-012628d5e3c8.png">

zulhfreelancer:
Here is my tests @jackycute —

Test 1

![Screen Shot 2020-02-07 at 10 13 28 AM](https://user-images.githubusercontent.com/3416976/73995090-9ad62600-4992-11ea-94e9-72aca79432d8.png)

![Screen Shot 2020-02-07 at 10 13 50 AM](https://user-images.githubusercontent.com/3416976/73995098-a1fd3400-4992-11ea-9dab-c397160f310c.png)

---

Test 2

![Screen Shot 2020-02-07 at 10 13 05 AM](https://user-images.githubusercontent.com/3416976/73995129-b5100400-4992-11ea-8b0f-142b8ab83aff.png)

![Screen Shot 2020-02-07 at 10 12 53 AM](https://user-images.githubusercontent.com/3416976/73995151-bfca9900-4992-11ea-9c23-a518e9839c98.png)

jackycute:
@zulhfreelancer I see that now.
[...]

PR #1416, opened 2020-02-26
Fix some issues after code refactoring

a60814billy:
1. fix server throw TypeError when user not sign-in
2. fix issue when shortId not found on publish link


jackycute:
this should fix #1402

jackycute:
LGTM, thanks @a60814billy

PR #1417, opened 2020-02-26
Support multi-row and row span table syntax

a60814billy:
This PR support two new table syntax, one is multi-row syntax that use `\` to merge multiple row into one row. another is row-span syntax that can merge multiple column into one colume (like html rowspan attribute)

1. multi-row syntax example 
```markdown
|   Markdown   | Rendered HTML |
|--------------|---------------|
|    *Italic*  | *Italic*      | \
|              |               |
|    - Item 1  | - Item 1      | \
|    - Item 2  | - Item 2      |
|    ```python | ```python       \
|    .1 + .2   | .1 + .2         \
|    ```       | ```           |
```

2. row span syntax
```
Stage | Direct Products | ATP Yields
----: | --------------: | ---------:
Glycolysis | 2 ATP ||
^^ | 2 NADH | 3--5 ATP |
Pyruvaye oxidation | 2 NADH | 5 ATP |
Citric acid cycle | 2 ATP ||
^^ | 6 NADH | 15 ATP |
^^ | 2 FADH2 | 3 ATP |
**30--32** ATP |||
[Net ATP yields per hexose]
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11657,6 +11657,28 @@
       "resolved": "https://registry.npmjs.org/markdown-it-mathjax/-/markdown-it-mathjax-2.0.0.tgz",
       "integrity": "sha1-ritPTFxxmgP55HXGZPeyaFIx2ek="
     },
+    "markdown-it-multimd-table": {
+      "version": "4.0.1",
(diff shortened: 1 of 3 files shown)


Yukaii:
Nice! But found that our existing table auto format would break this syntax

a60814billy:
Yap, it's a known issue, maybe we can add preference options to turn off table auto formatting.

PR #1419, opened 2020-02-26
fix: catch viz.js render async error and recreate Viz instance

### LINKED ISSUE #1374: Graphviz does not work in develop branch
bug
How to reproduce the error:
 1. Create new note
 2. Open "Both" view (Ctrl+Alt+B)
 3. Paste:
```graphviz
digraph a {
aaa
	
}
```
 4. Place cursor just after "aaa" and start typing characters: "-> b".
Just after you type character "-" graph stops to render (because of syntax error) and it will not begin to render after syntax error is fixed (after typing "-> b").

Upgrading viz.js (4531d96) may have caused this problem.

jackycute:
This PR fixes #1374 

- viz.js return Promise since version 2.0.0 and we didn't catch the error properly
- on catching error, we should recreate Viz instance to clean up the error, see more: https://github.com/mdaines/viz.js/wiki/Caveats

### DIFF
--- a/public/js/extra.js
+++ b/public/js/extra.js
@@ -31,7 +31,7 @@ require('prismjs/components/prism-gherkin')
 
 require('./lib/common/login')
 require('../vendor/md-toc')
-const viz = new window.Viz()
+let viz = new window.Viz()
 const plantumlEncoder = require('plantuml-encoder')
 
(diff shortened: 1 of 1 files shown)


jackycute:
Thanks @a60814billy

PR #1433, opened 2020-03-02
Disable HSTS IncludeSubdomain, to prevent unexpected behavior

a60814billy:
This is breaking change !!!
This PR change HSTS IncludeSubdomain setting to false.
In default setting, its will force subdomain use HSTS, but it's breaking the site that hosted in subdomain and didn't has SSL certificate.


jackycute:
Thanks @a60814billy

PR #1434, opened 2020-03-02
Change default anonymous setting for more privacy

a60814billy:
This is BREAKING CHANGE!!!!
To prevent any user can create guest note by default.

change allowAnonymous to `false` and setup
allowAnonymousEdits and allowAnonymousViews to `true`

more detail: https://hackmd.io/c/codimd-documentation/%2F%40codimd%2Fnote-permission#Default-Behavior


jackycute:
Thanks @a60814billy

PR #1435, opened 2020-03-02
Remove old config and environment

a60814billy:
BREAKING CHANGE: remove old config and environment


jackycute:
Thanks @a60814billy

PR #1443, opened 2020-03-05
Fix Wikipedia link in 2.0.0 release notes

Lin-Buo-Ren:
Fixes the problem where the link opens the same release notes page.

Signed-off-by: 林博仁(Buo-ren, Lin) <Buo.Ren.Lin@gmail.com>


jackycute:
LGTM, thanks @Lin-Buo-Ren

PR #1459, opened 2020-03-16
Improve version checker behavior

a60814billy:
This PR' is to improve version checker behavior.

1. setup timeout to 1000ms: if server's network is unstable, may cause response too slow.
2. change logger type from warn --> info, error --> warn


lib/web/middleware/checkVersion.js
```diff
@@ -30,11 +30,12 @@ async function checkVersion (ctx) {
     const { statusCode, body: data } = await rp({
       url: `${VERSION_CHECK_ENDPOINT}?v=${config.version}`,
       method: 'GET',
-      json: true
+      json: true,
+      timeout: 1000
```
a60814billy:
```suggestion
      timeout: 3000
```

PR #1463, opened 2020-03-27
Simple image lightbox

Yukaii:
codimd-image-lightbox

### TODOs

- [x] image zoom / drag
- [x] fix scroll behavior in published mode
- [x] customize markdown-it image renderer
- [x] dark
- [x] keyboard navigation

### Future enhancements (not in this PR)

- zoom scale transform origin(https://stackoverflow.com/a/15296290https://stackoverflow.com/a/15296290)
  - https://www.sitepoint.com/community/t/javascript-css3-mousewheel-zoom-origin-problem/221456/4
- touch support
- fit image to screen if image is oversize

### DIFF
--- a/public/js/extra.js
+++ b/public/js/extra.js
@@ -21,6 +21,7 @@ import {
   deserializeParamAttributeFromElement
 } from './lib/markdown/utils'
 import { renderFretBoard } from './lib/renderer/fretboard/fretboard'
+import './lib/renderer/lightbox'
 
 import markdownit from 'markdown-it'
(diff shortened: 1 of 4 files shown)


jackycute:
Good job, there are some points I want to mention:

- the zoom in/out should center from the cursor position, currently it zooms only from top left
- the dragging area is too small, it should cover the whole lightbox instead of the image itself, which might accidentally select the image of click out the lightbox
- I couldn't exit the lightbox with ESC key

PR #1470, opened 2020-04-01
Fix check for creating free url notes

ChrisMacNaughton:
The current check requires that both `config.allowAnonymous` is true,
and a user is logged in. This change modifies the check to allow either
of these to be true, allowing anonymous access OR logged in access
when creating free url notes

Signed-off-by: Chris MacNaughton <chmacnaughton@gmail.com>

### DIFF
--- a/lib/note/index.js
+++ b/lib/note/index.js
@@ -34,7 +34,9 @@ async function getNoteById (noteId, { includeUser } = { includeUser: false }) {
 }
 
 async function createNote (userId, noteAlias) {
-  if (!config.allowAnonymous && !userId) {
+  // The following check will throw an error if:
+  // config.allowAnonymous is False OR userId is not set
+  if (!config.allowAnonymous || !userId) {
     throw new Error('can not create note')
   }
 
--- a/lib/note/index.js
+++ b/lib/note/index.js
@@ -36,7 +36,7 @@ async function getNoteById (noteId, { includeUser } = { includeUser: false }) {
 async function createNote (userId, noteAlias) {
   // The following check will throw an error if:
   // config.allowAnonymous is False OR userId is not set
-  if (!config.allowAnonymous || !userId) {
+  if !(config.allowAnonymous || userId) {
     throw new Error('can not create note')
   }
 


ChrisMacNaughton:
It looks like the PR Travis build did pass, and somehow failed to communicate the state with Github?

jackycute:
Hi @ChrisMacNaughton,
Thanks for the PR.
However, I think the current logic is expected, just like in below table:

| create note            | user id: true | user id: false |
| ---------------------- | ------------- | -------------- |
| Allow anonymous: true  | true          | true           |
| Allow anonymous: false | true          | false          |

your proposal is:

| create note            | user id: true | user id: false |
| ---------------------- | ------------- | -------------- |
| Allow anonymous: true  | true          | false          |
| Allow anonymous: false | false         | false          |

ChrisMacNaughton:
The new logic may be incorrect, I'll have more through on it; the issue I'm trying to solve is:

```
codimd_1    | 2020-04-06T07:58:31.181Z info: 	deserializeUser: e7f5d641-1678-4a62-9efa-8ab17ce38fde
codimd_1    | Error: can not create note
codimd_1    |     at createNote (/home/hackmd/app/lib/note/index.js:38:11)
codimd_1    |     at showNote (/home/hackmd/app/lib/note/index.js:65:18)
codimd_1    |     at <anonymous>
codimd_1    | 2020-04-06T07:58:31.252Z info: 	172.20.0.1 - - [06/Apr/2020:07:58:31 +0000] "GET /test HTTP/1.1" 500 148 "-" "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:75.0) Gecko/20100101 Firefox/75.0"
```

ChrisMacNaughton:
Ok, I think this PR is incorrect, but I still think that the issue is present, as an example:

```
userid=true
anon=false
!anon && !userid == false
anon || userid == true
```

(In essence (!bool && !bool) != (bool || bool))

so, I think that this change should modify it to be 

```
if !(config.allowAnonymous || userId) {
    throw new Error('can not create note')
}
```

jackycute:
> The new logic may be incorrect, I'll have more through on it; the issue I'm trying to solve is:
> 
> ```
> codimd_1    | 2020-04-06T07:58:31.181Z info: 	deserializeUser: e7f5d641-1678-4a62-9efa-8ab17ce38fde
> codimd_1    | Error: can not create note
> codimd_1    |     at createNote (/home/hackmd/app/lib/note/index.js:38:11)
> codimd_1    |     at showNote (/home/hackmd/app/lib/note/index.js:65:18)
> codimd_1    |     at <anonymous>
> codimd_1    | 2020-04-06T07:58:31.252Z info: 	172.20.0.1 - - [06/Apr/2020:07:58:31 +0000] "GET /test HTTP/1.1" 500 148 "-" "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:75.0) Gecko/20100101 Firefox/75.0"
> ```

Hi, which version are you using?
I believe we merged a related PR https://github.com/hackmdio/codimd/pull/1460 and it's not been released yet.

ChrisMacNaughton:
```
$ cat /home/hackmd/app/lib/note/index.js
--SNIP--
async function createNote (userId, noteAlias) {
  if (!config.allowAnonymous && !!userId) {
    throw new Error('can not create note')
  }
```

It looks like that fix isn't even on the latest docker build, I'm using nabo.codimd.dev/hackmdio/hackmd:latest

ChrisMacNaughton:
That said, (!bool && !bool) is still not the same as !(bool || bool), so this PR may be more correct

jackycute:
`nabo.codimd.dev/hackmdio/hackmd:latest` is at version 2.0.0.
The PR #1460 is only available in the develop branch now (means not been released yet.)

> That said, (!bool && !bool) is still not the same as !(bool || bool), so this PR may be more correct

I'm sorry but I need to point out `(!bool && !bool)` is the same as `!(bool || bool)`, according to [De Morgan's laws](https://en.wikipedia.org/wiki/De_Morgan%27s_laws).

ChrisMacNaughton:
> `nabo.codimd.dev/hackmdio/hackmd:latest` is at version 2.0.0.
> The PR #1460 is only available in the develop branch now (means not been released yet.)
> 
> > That said, (!bool && !bool) is still not the same as !(bool || bool), so this PR may be more correct
> 
> I'm sorry but I need to point out `(!bool && !bool)` is the same as `!(bool || bool)`, according to [De Morgan's laws](https://en.wikipedia.org/wiki/De_Morgan%27s_laws).

I really shouldn't try thinking logically before enough coffee; what I mean was:

`(!bool && !bool)`isn't the same as `!(bool || bool)`, as can be seen with this bit of javascript:

```javascript
(!true && !false) != !(true || false)
```
The point of the above statement being that the code is currently checking (NOT anonymous AND NOT logged in), where either of those being false will raise the exception, where we should be checking for NOT (anonymous OR logged in), at least, as far as I can tell.

The [change linked above](1460) changes from (NOT anonymous AND NOT NOT logged in), which requires Anonymous editing to be enabled, and a user to not be logged in to the current state (Anonymous allowed and logged in), but doesn't allow a logged in user to create named notes without anonymous editing enabled.

ChrisMacNaughton:
Sorry for the noise, my brain has failed me this morning

jackycute:
It's ok, we all need to wake up sometimes.

PR #1473, opened 2020-04-05
fix: cannot upload image via imgur

a60814billy:
The code in `@hackmd/imgur` has some issue that doesn't require `q` properly


jackycute:
LGTM, thanks @a60814billy

Issue #1454, opened 2020-03-13
docker image use unmaintained debian jessie


philpep:
Hi,

The docker image uses an old version of debian which does no longer receive security updates.

```
% docker run --rm -it nabo.codimd.dev/hackmdio/hackmd:2.0.0 cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 8 (jessie)"
NAME="Debian GNU/Linux"
VERSION_ID="8"
VERSION="8 (jessie)"
ID=debian
HOME_URL="http://www.debian.org/"
SUPPORT_URL="http://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

a60814billy:
Hi @philpep,

Thanks for your information, the Debian jessie is supported to June 30 2020. We already plan updating to next LTS version along with node.js 12.

philpep:
Hi @a60814billy , ah indeed jessie is still supported, good to hear update is planned. I'm closing the issue.

Issue #1445, opened 2020-03-06
Can't build with node LTS 12.16.1


pkrasicki:
Is current LTS version not supported? The [docs](https://hackmd.io/c/codimd-documentation/%2Fs%2Fcodimd-manual-deployment) only mention node versions 8-11. I get this error:

```
./bin/setup
(...)

gyp ERR! build error 
gyp ERR! stack Error: `make` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/usr/local/lib/node_modules/npm/node_modules/node-gyp/lib/build.js:194:23)
gyp ERR! stack     at ChildProcess.emit (events.js:311:20)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:275:12)
gyp ERR! System Linux 4.19.0-8-amd64
gyp ERR! command "/usr/local/bin/node" "/usr/local/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
gyp ERR! cwd /home/user/codimd/node_modules/scrypt
gyp ERR! node -v v12.16.1
gyp ERR! node-gyp -v v5.0.5
gyp ERR! not ok 
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.2.11 (node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.2.11: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! scrypt@6.0.3 install: `node-gyp rebuild`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the scrypt@6.0.3 install script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

a60814billy:
Hi @pkrasicki,
Yes, it currently doesn't support node.js version 12. We're working on it.

a60814billy:
Currently, it is supposed to be Node.JS 10~16. This issue may have been resolved. If it continues to occur, feel free to reopen it.

PR #1474, opened 2020-04-08
Replace scrypt with scrypt-kdf

binotaliu:
This PR replaced the unmaintained library `node-scrypt` with a replacement `script-kdf`.

There is a native API for scrypt support from Node.js 10+, and **Node.js 8 has reached it's EOL on Dec 2019.** I think it's safe to say we can now switch to use the native API. The node version required in `packages.json` has been changed to `>= 10, < 13`, but need more work to check if it is fully functioned in Node 12.

As I changed the dependencies, following changes has been made:

 - The setter for `User.password` are now removed. As `scrypt-kdf` can only generate hashes asynchronized.
 - Added a new async method `User.hashPassword`. We need to manually call this method when setting password.
 - `User.prototype.verifyPassword` has been made async.

Besides, this means we should now be able to run CodiMD on Node 12.

Related to issue #1445, #1454

### DIFF
--- a/lib/models/user.js
+++ b/lib/models/user.js
@@ -57,6 +57,18 @@ module.exports = function (sequelize, DataTypes) {
     return false
   }
 
+  User.addHook('beforeCreate', async function (user) {
+    // only do hash when password is presented
+    if (user.password) {
+      user.password = await User.hashPassword(user.password)
+    }
+  })
+  User.addHook('beforeUpdate', async function (user) {
+    if (user.changed('password')) {
+      user.password = await User.hashPassword(user.password)
+    }
+  })
+
   User.associate = function (models) {
     User.hasMany(models.Note, {
       foreignKey: 'ownerId',
--- a/lib/models/user.js
+++ b/lib/models/user.js
@@ -1,7 +1,7 @@
 'use strict'
 // external modules
 var Sequelize = require('sequelize')
-var scrypt = require('scrypt')
+var Scrypt = require('scrypt-kdf')
 
 // core
 var logger = require('../logger')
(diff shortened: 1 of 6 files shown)


Yukaii:
Hi @binotaliu, could you please put the hash function inside sequelize hooks? Instead of calling the hash function every time, I prefer to let the model handle itself. Prevent we forget this when the code grows.

There're [two hooks](https://sequelize.org/master/manual/hooks.html#declaring-hooks): `beforeCreate` and `beforeUpdate` that you can use.

For the `beforeCreate` hook, since the password is not a required field, only save the hashed password when the password is supplied.

For the `beforeUpdate` hook, you can use [`#changed`](https://sequelize.org/master/class/lib/model.js~Model.html#instance-method-changed) method to determine whether the password field is updated or not.

Overall this PR looks awesome, thank you for the contribution! 😻

binotaliu:
@Yukaii Use hook is cool but this will make data inconsistent. The data you passed into model is a plain text password, once you save it and read the same property again, it become hashed.

```javascript
const me = User.build({})

me.email = 'me@example.com'
me.password = 'hunter2'

await me.save()

console.log(me.password) // what's the output?
```

Yukaii:
@binotaliu Yes, the data is inconsistent, but that's we expect. Current `setter` based implementation does the same. Once we set the password, we should not access the plain text password anymore.

binotaliu:
I still keep the `hashPassword` method public on `User`.

Two hooks, `beforeCreate` and `beforeUpdate`, has been added. Though, AFAIK, there is no way user can change their password right now.

There will be a known bug that user can't change their password to the hashed string that stored in the database because `user.changed('password')` will return `false`.

lib/models/user.js
```diff
@@ -41,22 +41,38 @@ module.exports = function (sequelize, DataTypes) {
...
-    } else {
-      return false
     }
+
+    return false
   }
 
+  User.addHook('beforeCreate', async function (user) {
+    // only do hash when password is presented
+    if (!user.password) {
```
Yukaii, line 62: Currently, password is an optional field in CodiMD, could you modified as follows:

```suggestion
    if (user.password) {
	  user.password = await User.hashPassword(user.password)
    }
```

lib/models/user.js
```diff
@@ -41,22 +41,38 @@ module.exports = function (sequelize, DataTypes) {
...
+      return
+    }
+
+    user.password = await User.hashPassword(user.password)
+  })
+  User.addHook('beforeUpdate', async function (user) {
+    if (!user.changed('password')) {
```
Yukaii, line 69: Same as above, the password is optional.

Yukaii:
@binotaliu The changes look good to me now 😄 There are conflicts in `package.json`, please resolve them. 👍

binotaliu:
@Yukaii I have rebased the branch onto develop. the conflicts has been resolved.

Yukaii:
Thank you @binotaliu

PR #1481, opened 2020-04-10
Set proper "lang" attribute on <html> and #doc

binotaliu:
This pull request fix the issue that `lang` attribute are always set to "en" no matter which language the user is using.
Also, note language might be different than user interface. You can not set the language of the note by a YAML meta `lang`. This is not a new meta tag. It already exists on the code base: https://github.com/hackmdio/codimd/blob/e72bcfe0eab7084c6ade055db24a73d634f3d6f6/public/js/extra.js#L731-L742

But this only affect to exported HTML. I changed the `index.js` so it would set lang attribute on `#doc`.

Setting up a proper `lang` attribute is very necceserry for a11y. Since some screen reader like VoiceOver on iOS/macOS will relying on lang attribute to detect content language.


Yukaii:
`getLocale()` would fail for simplified and traditional Chinese now, since their proper language code should be `zh-Hans` and `zh-Hant` but file names are `zh-CN.json` and `zh-TW.json` respectively. Or we can just rename these two files. What do you think? @jackycute 

ref: https://www.w3schools.com/tags/ref_language_codes.asp

binotaliu:
@yukaii `zh-TW` and `zh-CN` is valid [per BCP 47, Section 2.2.4 Region subtag](https://tools.ietf.org/html/bcp47#section-2.2)

https://www.w3.org/TR/html51/dom.html#element-attrdef-global-lang
https://tools.ietf.org/html/bcp47#section-2.2

jackycute (approved):

jackycute:
Thanks!

PR #1490, opened 2020-04-20
Fix GitHub's avatar URL

### LINKED ISSUE #1489: GitHub avatar invalid URL
## What

Noticed that after authenticating with github, the avatar URL is a little malformed.

```
https://avatars3.githubusercontent.com/u/2418945v=4?s=96
```

Notice the `v=4` just after ID is effectively part of the ID, which should not be the case.

It's caused by this piece of code:

https://github.com/hackmdio/codimd/blob/d7cc95129d98fe81adf15dee973a434f9a739476/lib/models/user.js#L117-L122

## Potential solution

```javascript
const u = new URL(profile.photos[0].value);
if (bigger) { u.searchParams.set('s', 400) } else { u.searchParams.set('s', 96) }
photo = u.toString();
```

Or more human example:

```javascript
const u = new URL('https://avatars3.githubusercontent.com/u/2418945?s=60&v=4');
u.searchParams.set('s', 96);
console.log(u.toString());
```

paroxp:
## What

At the moment, the URL is being composed and modified with the use of
string composition.

This causes issues, if the URL returned by GitHub slightly differs from
the time developer initially had a look into it.

In our case, the URL from GitHub has two query parameters in it, whilst
the codebase only expected one.

This change will take all of these parameters and only set the one we
care about, whilst leaving others intact and carry on with the full URL.

Fixes #1489

## How to review

- Run locally with GitHub OAuth enabled
- Login with GitHub
- See your avatar displayed properly
- *optionally* Follow the same steps on the `develop` branch to see difference

### DIFF
--- a/lib/models/user.js
+++ b/lib/models/user.js
@@ -115,10 +115,11 @@ module.exports = function (sequelize, DataTypes) {
         else photo += '?size=bigger'
         break
       case 'github':
-        if (profile.photos && profile.photos[0]) photo = profile.photos[0].value.replace('?', '')
-        else photo = 'https://avatars.githubusercontent.com/u/' + profile.id
-        if (bigger) photo += '?s=400'
-        else photo += '?s=96'
+        const photoURL = new URL(profile.photos && profile.photos[0]
+          ? profile.photos[0].value
+          : `https://avatars.githubusercontent.com/u/${profile.id}`)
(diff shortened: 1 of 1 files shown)


Yukaii:
Test on my local machine and it works well. Thank you @paroxp

PR #1501, opened 2020-04-29
Use array for tags when available (close #1496)

### LINKED ISSUE #1496: Tags in YAML metadata can't handle arrays
I can see CodiMD [can only handle a comma-separated string of tags](https://github.com/hackmdio/codimd/blob/develop/lib/models/note.js#L371), so using array notation doesn't work:

```
tags: [tag1, tag2, tag3]
tags:
  - tag1
  - tag2
  - tag3
```

That is valid YAML syntax and such an enhancement would greatly benefit interoperability with other Markdown processing software. Thanks!!!

daniele-athome:
This patch introduces metadata tags in array format.

Metadata tags can be done by either using strings (like before):

```
---
tags: tag1, tag2, tag3
---
```

And, with this patch, also using arrays, in any YAML supported format:

```
---
tags:
  - tag1
  - tag2
  - tag3
---
```

```
---
tags: [tag1, tag2, tag3]
---
```

This will make it more compatible with other software using structured types in the metadata section (I use CodiMD with other markdown software).

### DIFF
--- a/lib/models/note.js
+++ b/lib/models/note.js
@@ -367,8 +367,13 @@ module.exports = function (sequelize, DataTypes) {
   Note.extractNoteTags = function (meta, $) {
     var tags = []
     var rawtags = []
+    var metaTags
     if (meta.tags && (typeof meta.tags === 'string' || typeof meta.tags === 'number')) {
-      var metaTags = ('' + meta.tags).split(',')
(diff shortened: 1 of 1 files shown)


daniele-athome:
Deprecated by #1505.

Issue #90, opened 2016-03-03
Saving to local filesystem


joshlk:
Hi,

Great project! I've been looking for something like this for a while. I would like to setup my own private site and add functionality so I can save and load markdown files to a local filesystem - in my own hackdmd fork (happy to merge if desired). Im a programmer but have not experience in node.js/javascript. How hard would this be? Could you please direct me to the right areas in the code for me to look at.

Thanks

jackycute:
Hi, Thanks for loving this project!
node.js/javascript are not hard to learn but hard to be good.

Currently, all notes will store in database which is totally different from filesystem.
You can import and export markdown files, we have those features.
But you can't browse all your filesystem or markdown files in some kind of tree view right now.

Because we have to manage permission and user auth problems, the db is necessary.
But we can instead use sqlite to manage that and use filesystem to store notes.

Anyway, these will not be an easy work because these will need ORM the db which I'm working on.
And implement the filesystem tree view etc.

It's really great to hear that you want to help.
You can start from the server side will be easier.
https://github.com/hackmdio/hackmd/tree/master/lib

I might be to help after some other works and issues be done. :smile:

SISheogorath:
I'm not sure how usual and how needed this feature is.

It has a high potential to cause security issues (as always when you provide access to the server side filesystem) and I don't think that there are many users. But maybe I'm wrong, so what do you think?

ccoenen:
(I am assuming you mean write/read from the server's disk, _not_ from the client's disk. In case, I misunderstood, just skip my comment.)

Writing the notes out to the filesystem is not hard, In case you need pointers, there's a Pull-Request exporting to git: Pull Request #303 / the line starting the [full export to disk](https://github.com/hackmdio/hackmd/pull/303/files#diff-f7292c806ccb90e66b57ced437cb1c28R22). Importing might not be as easy, I'm not sure if you need to satisfy any other constraints other than "write to notes table".

To be quite honest, I'm also not yet convinced this requirement is very common.

markusressel:
I'm currently using [MkDocs](http://www.mkdocs.org/) as a private Wiki.
MkDocs is a static site generator though and they have no intention to implement an online editor for it.
There are almost no alternatives to editing the files directly on the filesystem (there is [Pendulum](https://github.com/titpetric/pendulum) but it has very limited capabilities compared to HackMD).

[...]

PR #1504, opened 2020-04-30
Write note to filesystem if file exists

daniele-athome:
This patch writes note content back to the filesystem if a file exists in the public docs path. This can be a first step towards the real objective of #90 maybe?

Please note that this doesn't address the fact that notes loaded from docsPath are set with a null user owner ID so they are not actually editable. You need to update the row in the database for that. I believe that part should be addressed in another way (I was thinking something like per-user docsPath)

### DIFF
--- a/lib/models/note.js
+++ b/lib/models/note.js
@@ -129,6 +129,13 @@ module.exports = function (sequelize, DataTypes) {
             return resolve(note)
           })
         })
+      },
+      afterUpdate: function (note, options, callback) {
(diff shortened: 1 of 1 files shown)


jackycute:
Hi @daniele-athome,
I think this is great! It will be better if there is a config to enable/disable this feature.

daniele-athome:
I'm sorry but I stopped using CodiMD :cry:

jackycute:
> I'm sorry but I stopped using CodiMD 😢

It's ok 😃 , hope you doing good.

PR #1520, opened 2020-05-18
chore: add .nvmrc to fix node version

a60814billy:
We're use node.js dubnium lts (10.20.1) for docker image, therefor add .nvmrc to specify node.js version


Yukaii:
Thank you @a60814billy

PR #1522, opened 2020-05-19
fix: replace btn-social with btn-login-method

### LINKED ISSUE #1491: Consider removing the "btn-social" class for custom OAuth2 providers (and potentially others)
I will preface this issue by saying that this problem is caused by using an ad-blocker and thus might be out of scope for CodiMD (and therefore I am not creating a pull request directly).

The "btn-social" class is blocked by the "Fanboy's Social" and "Fanboy's Annoyance" filter lists which are available in e.g. uBlock origin by default (although I am not sure if they are enabled by default). Therefore the "Sign in via ..." button is not visible.

While this is most likely intended behavior for e.g. the Facebook, Twitter and Google buttons (and potentially some others) and people who enable these filter lists probably won't miss them, this also blocks self-hosted solutions like custom OAuth2 providers.

It should however be noted that this changes the style of the button (the button on the top in the picture is with the `btn-social` class, the bottom one without the class).

Screenshot from 2020-04-20 18-40-31

a60814billy:
avoid to use btn-social, some AD block will delete
the DOM with btn-social class.

fixed #1491

### DIFF
--- a/public/css/cover.css
+++ b/public/css/cover.css
@@ -397,7 +397,23 @@ select {
     color: #fff;
 }
 
-a.btn.btn-social > i.oauth-icon {
+/* add btn-login-method replaced of btn-social, to avoid ad block delete the login button */
+.btn-login-method {
+  padding-left: 61px;
+  position: relative;
+  text-align: left;
+  white-space: nowrap;
+  overflow: hidden;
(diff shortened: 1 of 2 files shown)


Yukaii:
Thank you @a60814billy

PR #1530, opened 2020-05-30
fix: improve image upload to filesystem may caused app crash

a60814billy:
This PR fixes users using filesystem as image storage and not setup proper permission may caused application crash.

1. always uploads file to os temporary path (formidable library default behavior)
2. In filesystem image upload module, move temp file into upload folder and if has any problem, return error properly.


jackycute:
LGTM, thanks @a60814billy

Issue #368, opened 2017-02-22
table of contents tag only displays 3 levels of headline


mxmilkiib:
While I can see benefits for this in some situations, I have the first H1 as the document title so my 'third level' document body is ####, it would be handy if headers of level 4 (and possibly 5) could be displayed via [TOC] or maybe a new tag.

a60814billy:
Thanks @milkmiruku 
Maybe we can add option to TOC tag,
like `[TOC level=4] ` to set TOC level ?

ccoenen:
@milkmiruku if you only need the title for list/overview, you could also add it to the yaml metadata at the top like this

```yaml
---
title: This is my awesome document. There are many like it, but this one is mine!
---
```
see https://hackmd.io/yaml-metadata for more detailed info.

milkmiruku:
@ccoenen thanks, it's more I'd like to use h1 for the document name/title. An alternative idea could be an option to shunt the TOC so it ignores h1 (if there is only one)?

jackycute:
@milkmiruku currently, we don't have an option for customizing TOC ranges.
But maybe this would be nice to have.

jackycute:
I found a good example for the YAML metadata options.

https://shd101wyy.github.io/markdown-preview-enhanced/#/toc?id=toc-and-sidebar-toc-configuration

salim-b:
> I found a good example for the YAML metadata options.
> 
> https://shd101wyy.github.io/markdown-preview-enhanced/#/toc?id=toc-and-sidebar-toc-configuration

An option to set the ToC depth would be really helpful!

Maybe you'd like to just use the same syntax as [R Markdown does](https://bookdown.org/yihui/rmarkdown/html-document.html#table-of-contents)? It looks like this:

```rmd
---
title: "Habits"
output:
  html_document:
    toc: true
    toc_depth: 2
---
```

Then we could set the ToC _separately for each export format_. Of course that would imply some additional implementation work, but since both HackMD/CodiMD _and_ R Markdown/Knitr are relying on Pandoc to convert the Markdown, that would really make sense I guess (the above YAML keys correspond to [Pandoc's CLI options](https://pandoc.org/MANUAL.html#general-writer-options) `--toc` and `--toc-depth`). 🙃

jackycute:
We supported this by https://github.com/hackmdio/codimd/pull/1532
You can use YAML metadata or use `[ToC maxLevel=4]` to specify toc levels.

PR #1306, opened 2019-10-20
Configurable default break style

Yukaii:
resolves #1212 

### Changes

This PR introduce a new configuration entry let you follow CommonMark spec break behavior:

|Set `config.defaultUseHardbreak` to `true` <br> (Current CodiMD implementation) | Set `config.defaultUseHardbreak` to `false` <br> (CommonMark)|
|--|--|
|Screen Shot 2019-10-20 at 2 22 00 PM|Screen Shot 2019-10-20 at 2 22 39 PM|

Notice that we keep the default `defaultUseHardbreak` value to `true` for preserving the current behavior.

### TODOs

- [ ] Update documentation

### Ref

- https://github.com/jdhoek/codimd/commit/c12a8d2fddf496966f64ccabc688f3a09320ea90 in https://github.com/hackmdio/codimd/issues/1212#issuecomment-500800293

PR #1532, opened 2020-06-01
Set TOC depth freely for every note by using YAML metadata or an option within `[toc]`

zergar:
This commit allows one to set the depth of the table of contents freely by means of YAML metadata or an option directly within the `[toc]` tag (e.g. `[toc maxLevel=4]`. The default level can also be set by means of the environment variable `CMD_DEFAULT_TOC_DEPTH`, this variable defaults to 3. The toc depth of the automatically generated toc can only be set by setting the level in the YAML or below.

The precedence of the toc depth is as follows (highest to lowest): `[toc maxLevel=4]` > YAML metadata > `CMD_DEFAULT_TOC_DEPTH` > 3 (default value)

codimd-toc-example-customtoc

## Optional Features
+ [x] It might be nice to set a default depth with an environment variable, though I'm unsure how to access them from the frontend (as I found no example)

### DIFF
--- a/lib/config/environment.js
+++ b/lib/config/environment.js
@@ -148,5 +148,5 @@ module.exports = {
   defaultUseHardbreak: toBooleanConfig(process.env.CMD_DEFAULT_USE_HARD_BREAK),
   linkifyHeaderStyle: process.env.CMD_LINKIFY_HEADER_STYLE,
   autoVersionCheck: toBooleanConfig(process.env.CMD_AUTO_VERSION_CHECK),
-  defaultTocDepth: toIntegerConfig(process.env.CMD_DEFAULT_TOC_DEPTH, 3)
+  defaultTocDepth: toIntegerConfig(process.env.CMD_DEFAULT_TOC_DEPTH)
 }
--- a/lib/config/environment.js
+++ b/lib/config/environment.js
@@ -148,5 +148,5 @@ module.exports = {
   defaultUseHardbreak: toBooleanConfig(process.env.CMD_DEFAULT_USE_HARD_BREAK),
   linkifyHeaderStyle: process.env.CMD_LINKIFY_HEADER_STYLE,
   autoVersionCheck: toBooleanConfig(process.env.CMD_AUTO_VERSION_CHECK),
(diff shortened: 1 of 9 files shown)


zergar:
could resolve #368

Yukaii:
> It might be nice to set a default depth with an environment variable, though I'm unsure how to access them from the frontend (as I found no example)

For now, you can take a look at PR 1306 to see how it pass environment variable config to frontend: https://github.com/hackmdio/codimd/pull/1306

zergar:
> For now, you can take a look at PR 1306 to see how it pass environment variable config to frontend: #1306

Thank you very much for the tip, I've added the feature accordingly.

zergar:
to finish this up, I've added the option to set a custom TOC level within the toc tag, e.g. `[toc maxLevel=4]`. I've updated the PR description accordingly.

lib/config/environment.js
```diff
@@ -147,5 +147,6 @@ module.exports = {
   openID: toBooleanConfig(process.env.CMD_OPENID),
   defaultUseHardbreak: toBooleanConfig(process.env.CMD_DEFAULT_USE_HARD_BREAK),
   linkifyHeaderStyle: process.env.CMD_LINKIFY_HEADER_STYLE,
-  autoVersionCheck: toBooleanConfig(process.env.CMD_AUTO_VERSION_CHECK)
+  autoVersionCheck: toBooleanConfig(process.env.CMD_AUTO_VERSION_CHECK),
+  defaultTocDepth: toIntegerConfig(process.env.CMD_DEFAULT_TOC_DEPTH, 3)
```
Yukaii, line 151: The second argument is unnecessary.

```suggestion
  defaultTocDepth: toIntegerConfig(process.env.CMD_DEFAULT_TOC_DEPTH)
```

zergar: I'm approving this, but I'm not sure whether the toc depth isn't set anywhere else. A quick test though showed that somewhere a depth of 3 seems to be set.

zergar:
@jackycute does this look proper to you?

Issue #1545, opened 2020-06-27
Support YAML metadata image tag


schokotets:
In order to allow tools like [go-readability](https://github.com/go-shiori/go-readability) to parse codimd pages and retrieve an article image, the ability to set an image yaml value would be sensible.
Just like the description metadata tag, it should be a simple pass-through of the value to a corresponding html <meta> tag.

PR #1546, opened 2020-06-27
feature: pass-through yaml metadata image to html meta tag

schokotets:
This is referencing the issue 'Support YAML metadata image tag #1545'


Yukaii:
Thank you @schokotets

PR #1548, opened 2020-06-30
List-my-note API

JamesCamel:


lib/routes.js
```diff
@@ -70,6 +70,8 @@ appRouter.get('/s/:shortid/:action', response.publishNoteActions)
 appRouter.get('/p/:shortid', response.showPublishSlide)
 // publish slide actions
 appRouter.get('/p/:shortid/:action', response.publishSlideActions)
+// gey my note list
+appRouter.get('/myNotes', noteController.listMyNotes)
```
jackycute, line 74: Maybe we can use like `/api/notes` instead of root level path to avoid occupied more routes when introduce more APIs.

Yukaii: Yes we should do this. @JamesCamel Please change the route to `/api/notes/myNotes` 👍

lib/note/index.js
```diff
@@ -190,6 +189,53 @@ async function noteActions (req, res) {
   }
 }
 
+async function getMyNoteList (userid, callback) {
```
Yukaii, line 192:
```suggestion
async function getMyNoteList (userId, callback) {
```

lib/note/index.js
```diff
@@ -190,6 +189,53 @@ async function noteActions (req, res) {
...
+    where: {
+      ownerId: userid
+    }
+  })
+  if (!myNotes) {
+    return callback(null, null)
+  }
+  try {
+    const myNoteList = []
+    for (let i = 0; i < myNotes.length; i++) {
```
Yukaii, line 203:
```js
const myNoteList = myNotes.map(note => ({
        id: Note.encodeNoteId(note.id),
        text: note.title,
        tags: Note.parseNoteInfo(note.content).tags,
        createdAt: note.createdAt,
        lastchangeAt: note.lastchangeAt,
        shortId: note.shortid
}))
```

lib/routes.js
JamesCamel: Fixed.

lib/note/index.js
JamesCamel: Fixed.

lib/note/index.js
JamesCamel: Fixed.

PR #1552, opened 2020-07-07
Allow Sequelize CLI to use options set in config.json

### LINKED ISSUE #1396: Improve file base database config & documentation
When running `sequelize db:migrate`, sequelize-cli takes `.sequelizerc`'s config, but in our `.sequelizerc.example`:

```js
module.exports = {
  config: path.resolve('config.json'),
  'migrations-path': path.resolve('lib', 'migrations'),
  'models-path': path.resolve('lib', 'models'),
  url: process.env['CMD_DB_URL'] || config.dbURL
}
```

`config: path.resolve('config.json')` this line never works because sequelize `config.json` should be in the [following format](https://sequelize.org/master/manual/migrations.html#configuration):

```json
{
  "development": {
    "username": "root",
    "password": null,
    ...
  }
}
```

But in our `config.json.example`:

```json
  "development": {
    "db": {
      "username": "root",
      "password": null,
      ...
    }
  }
```

Developers need to set `dbURL` in config.json in order not to write it twice:

```json
{
  "development": {
    "dbURL": "postgres://localhost:5432/codimd",
    ...
  }
}
```

MHajoha:
Until now, the Sequelize CLI received options differently from the actual Sequelize instance created in `lib/models/index.js`. Aside from just being irritating, it prevents any database options from `config.json` from being used by the Sequelize CLI without duplicating them, as @Yukaii points out in #1396.

This PR has the Sequelize CLI use a `config.js` which uses `lib/config` to load the configuration just as the application does.

Fixes #1396
Signed-off-by: Maximilian Haye <mhajoha@gmail.com>

### DIFF
--- a/.sequelizerc.example
+++ b/.sequelizerc.example
@@ -2,8 +2,8 @@ const path = require('path')
 const config = require('./lib/config')
 
 module.exports = {
-  config: path.resolve('config.json'),
+  config: path.resolve('config.js'),
   'migrations-path': path.resolve('lib', 'migrations'),
   'models-path': path.resolve('lib', 'models'),
-  url: process.env['CMD_DB_URL'] || config.dbURL
+  url: config.dbURL
 }
--- a/config.js
+++ b/config.js
@@ -1,0 +1,3 @@
+const config = require('./lib/config')
+
+module.exports = config.db


MHajoha:
@a60814billy, roughly when will you be able to give this PR a review?

PR #1571, opened 2020-08-13
Allow specifying option for graphviz

### LINKED ISSUE #1568: GraphViz engine option
Having graphviz in markdown files is great, but being able to switch the rendering engine would be better, for example to select `circo` and have a nicer looking graph for nodes with many edges. The demo for [Viz.js](http://viz-js.com/) (which is the library I understood CodiMD uses for graphviz) shows the difference between engines, and the API has a `engine` [render option](https://github.com/mdaines/viz.js/wiki/API#render-options), so I suppose this should not be too hard to implement, and I could even have a go at it myself. The more important question is, how to add the engine option in Markdown?

We could add it just as another word?

~~~ markdown
``` graphviz circo
digraph {
}
```
~~~

Or as an argument, similar to [blockquotes](https://demo.codimd.org/s/features#Blockquote-Tags)?

~~~ markdown
``` graphviz [engine=circo]
digraph {
}
```
~~~

Yukaii:
fixes #1568 

### TODOs

- [x] Wait for https://github.com/hackmdio/CodeMirror/pull/8 then bump CodeMirror version to support graphviz syntax highlight

### Changes

Now you can specify [render options](https://github.com/mdaines/viz.js/wiki/API#render-options) as key="value" within braces:

~~~md
```graphviz {engine="circo"}
digraph hierarchy {

                nodesep=1.0 // increases the separation between nodes

                node [color=Red,fontname=Courier,shape=box] //All nodes will this shape and colour
                edge [color=Blue, style=dashed] //All the lines look like this

                Headteacher->{Deputy1 Deputy2 BusinessManager}
                Deputy1->{Teacher1 Teacher2}
                BusinessManager->ITManager
                {rank=same;ITManager Teacher1 Teacher2}  // Put them on the same level
}
```
~~~

codimd-graphviz-params

Syntax reference: pandoc's `fenced_code_attributes`

#### New graphviz syntax highlighting support

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -786,9 +786,9 @@
       "dev": true
     },
     "@hackmd/codemirror": {
-      "version": "5.49.5",
-      "resolved": "https://registry.npmjs.org/@hackmd/codemirror/-/codemirror-5.49.5.tgz",
-      "integrity": "sha512-9R/5wfex+Ik0L5TChqByf/SpwMTRXCrbiD0TB+R61CTfpRIv4LJp/khOxRFoZNvpFu6LXuejey+2gjD47PDS9g==",
+      "version": "5.49.8",
+      "resolved": "https://registry.npmjs.org/@hackmd/codemirror/-/codemirror-5.49.8.tgz",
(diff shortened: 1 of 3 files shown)


jackycute:
Thanks @Yukaii

Issue #1581, opened 2020-08-24
i18n: TOC menu has embedded English items


tamo:
TOC menu cannot be translated. e.g. "Expand all" and "Back to top" in public/js/extra.js

tamo:
Also, editor's placeholder text (in public/js/lib/editor/index.js) cannot be translated.

tamo:
Permissions (Freely, Editable, etc) are not translatable, either. (public/views/codimd/body.ejs)

tamo:
Status msgs (CONNECTED, ONLINE and OFFLINE in public/js/index.js) are not, either.

PR #1587, opened 2020-09-01
Move HTML-related code from JS to EJS to enable more i18n

tamo:
Don't manipulate HTML in JS but just show/hide existing elements.
This way, we can translate more.
See #1581


jackycute (approved):
Thanks @tamo

PR #1605, opened 2020-09-25
Fix matchInContainer false positives

tamo:
The function should match only the beginnings of lines.

For example, see this testcase:
```
:::spoiler
here is a :::
:::
:::
```
The last line should be completed, and the third should not.

Without this patch, the third line is completed and the last is not.

FYI, this patch is not perfect, either.
If the third line is ":::bogus", the parser won't treat it as the closer. But matchInContainer will.
So the right regexp would be /(^|\n):::\s*(success|info|warning|danger|spoiler)($|\s)/ but who cares

### DIFF
--- a/public/js/index.js
+++ b/public/js/index.js
@@ -3137,7 +3137,7 @@ function checkInContainerSyntax () {
 
 function matchInContainer (text) {
   var match
-  match = text.match(/:{3,}/g)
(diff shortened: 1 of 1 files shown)


Yukaii:
@tamo Commit 4002306 needs to be signed off to make DCO check passed. 👍

jackycute (approved):
LGTM

PR #1607, opened 2020-09-28
Feature/show shared note on history page

chouHong-Ming:
# New Feature

1. To display the team's shared note on the history page. If the user is already login, they can see the note that permission is set at freely, editable, limited, locked, protected, or they just can see the note that permission is set at freely, editable, locked.

2. In order to show the tags and search by the tags in the shared note section, to add a column to save the tags of the note in the "Note" table.

3. When the login user updating the note, every time loading the note, it will update the tags.

There is the built image that can be used to test and use : [https://hub.docker.com/r/chouhongming/codimd/tags](https://hub.docker.com/r/chouhongming/codimd/tags)


Yukaii:
@chouHong-Ming Could you please rebase onto the latest `develop` branch?

chouHong-Ming:
@Yukaii I already rebase onto the latest `develop` branch. But I can't pass DCO due to the rebase conflict and I can't change the email.

Yukaii:
@chouHong-Ming Sorry for the late reply. Can't you change email with `git config` command?

PR #1626, opened 2020-11-21
Add oauth2 authorization

joachimmathes:
Hi there,

the oauth2 provider is missing a basic authorization concept. That's why I propose this PR. The implementation is based on the saml provider:
https://github.com/hackmdio/codimd/blob/381b3ff78ec1569a142c9487f85756467618ef16/lib/auth/saml/index.js#L23

1. I added an attribute to specify the claim which is supposed to hold an array of strings of role names: `CMD_OAUTH2_ROLES_CLAIM`.
2. I added an attribute to specify a dedicated role which has to be included in the role claim of the ID token to become authorized.
3. Besides that I decided to allow another optional user profile attribute as `id`, which overrides `userProfileUsernameAttr`, if set: `userProfileIdAttr`.
  I introduced that attribute because a `username` might be unique, but nevertheless it is also prone to change. Imagine usernames which are built from parts of your first and lastname, either automatically or due to company guidelines. If people want to change their name, because of marriage or whatever reason else, the username no longer provides a stable ID. Thus, it's not only uniqueness but also immutability, which makes a good ID, e.g. a `uuid` provided by another claim

**Example configuration**
```
CMD_OAUTH2_ROLES_CLAIM = roles
CMD_OAUTH2_ACCESS_ROLE = role/codimd
CMD_OAUTH2_USER_PROFILE_ID_ATTR = user_uuid
```

Please note: I am no JavaScript developer. So, the code style of my modifications is merely an educated guess. :slightly_smiling_face:


jackycute:
Thanks @joachimmathes

PR #1636, opened 2020-12-22
Check upload image mime type

### LINKED ISSUE #1634: Unrestricted File Upload in uploadImage endpoint
This was also found during the hxp ctf. It seems to be active when `CMD_IMAGE_UPLOAD_TYPE` = `filesystem`.
 ### Credit:
@CherryWorm and his [writeup](https://ctftime.org/writeup/25438)

Steps to reproduce:
- Deploy using [docker-compose](https://hackmd.io/c/codimd-documentation/%2Fs%2Fcodimd-docker-deployment#Using-docker-compose-to-setup-CodiMD)
- run:
  ```sh
  $ touch test.js
  $ curl http://localhost:3000/uploadImage -F "image=@test.js"
  {"link":"/uploads/upload_529cb31201c94b38767a256ef625a6ab.js"}
  $ curl http://localhost:3000/uploads/upload_529cb31201c94b38767a256ef625a6ab.js -v
  ...
  Content-Type: application/javascript; charset=UTF-8
  ...
  ```
This makes it possible to upload arbitrary JS and HTML files which leads to XSS (CSP allows `script-src 'self'`).

Maybe you could check [`allowedUploadMimeTypes`](https://github.com/hackmdio/codimd/search?q=allowedUploadMimeTypes) server-side?

Yukaii:
Fixes #1634


nename0:
I don't think this quite fixes the problem. 

As far as I can trace it down [serve-static](https://github.com/expressjs/serve-static) uses this code to set the `Content-Type` on a response: 
https://github.com/pillarjs/send/blob/master/index.js#L833-L849
As you can see this only uses the file extension of the requested file.

So when you check the first 12 Bytes of the file for the file magic, that does not really matter for the `Content-Type` on a response.

Here is a proof of concept: [test.html.txt](https://github.com/hackmdio/codimd/files/5729659/test.html.txt)
Rename the file back to `test.html`
The first 12 Bytes of this file are directly copied from an valid png image.
However after that I inserted `<script>alert(1)</script>`.
Because of the first 12 Bytes [`image-type`](https://www.npmjs.com/package/image-type) will say, oh thats a png.
But when you serve this via [serve-static](https://github.com/expressjs/serve-static) it will see the `.html` extension and set `Content-Type: text/html; charset=UTF-8`.
Then the browser will start to interpret the result as html ignoring the first 12 Bytes (because browsers ignore most html errors) and then execute `alert(1)`.

Of course this won't work directly on codimd because of CSP (set to no inline script), but you see my point. A JS file could very likely be spoofed the same way.

nename0:
I just tried it on the deployment:
It works: https://codimd-bugfix-check-ima-f03las.herokuapp.com/uploads/upload_2a040560ebb65c71daf43ececec4ccdd.html
Now to circumvent the CSP you could use the method [from Alemmi](https://github.com/Alemmi/ctf-writeups/blob/main/hxpctf-2020/hackme/solution.md)

jackycute:
> Rename the file back to `test.html`

I don't think renaming it can be seemed as a step because user can't do it from client side.
If we need to prevent all the files serve from uploads folders then we need to put them behind some proxy or route and validate them on the fly.

nename0:
What user? The browser?

Just use this command:
`curl https://codimd-bugfix-check-ima-f03las.herokuapp.com/uploadImage -F "image=@test.html"`
It's not about client-side (what the browser can do) its about server-side (what can a malicious users do with your API).

jackycute:
How about we check the file extension that you try to upload? Will this work?

nename0:
Yes.
Exactly what i proposed [here](https://github.com/hackmdio/codimd/issues/1634#issuecomment-748896343)

jackycute:
And I think force to rename the image file extension according to the MIME type will also work.

nename0:
For consistency it would make sense to reuse the mime-type whitelist from here:
https://github.com/hackmdio/codimd/blob/381b3ff78ec1569a142c9487f85756467618ef16/lib/config/index.js#L177-L194

Yukaii:
Thanks for the help on this PR @nename0 

Could you still reproduce the issue on the latest [deployment](https://codimd-bugfix-check-ima-f03las.herokuapp.com/)?

nename0:
No I cannot reproduce, as now the file extensions are restricted and it is impossible to get a different Content-Type than `Content-Type: image/*` from the `/uploads/` Folder.
LGTM :+1:

PR #1650, opened 2021-01-21
Fix slide mode stored XSS

### LINKED ISSUE #1648: Stored XSS in slide mode (via reveal-markdown)
There is a (quite convoluted) stored XSS vulnerability in the slides feature:

 - The presentation's YAML options allow specifying JavaScript files as dependencies. Any JS file specified by the `dependency` option is loaded and executed (when viewed via the presentation mode, `/p/...`), subject to the server's CSP and their Content-Type header
 - We can use JSONP endpoints that bypass the CSP (e.g. on Vimeo or Slideshare, or on Disqus with a free API key) to run arbitrary existing JS functions _without arguments_
 - We use `Reveal.navigateRight()` followed by `RevealMarkdown.processSlides()` to inject HTML from a `div` element with a `data-markdown` attribute in the speaker notes into the DOM. The slide navigation is necessary to ensure that the speaker notes are outside of the DOM of the original slide before the second request comes in (duplicate calls to `processSlides` break the proof-of-concept).
 - Usually, the included markdown would be enclosed in a `<script type="text/template">` tag to ensure that nothing can escape into the DOM. However, `reveal-markdown.js` does not properly escape `</script>` tags (the check is case sensitive, but should at the very least be case insensitive):

```javascript
    // prevent script end tags in the content from interfering
    // with parsing
    content = content.replace(/<\/script>/g, SCRIPT_END_PLACEHOLDER)
```

 - Actual JS in this DOM injection will never be loaded, because it is assigned via `innerHTML`, but because the CSP includes so many different embed features, we can load [this jQuery templating code](https://a.disquscdn.com/1587400073/js/src/templates.js) from Disqus (it is used on the Disqus login page, fairly easy to find) that walks the entire DOM and renders jQuery template strings.
 - From the template, we can simply grab another script from anywhere (including another note) and `eval` it.

A full proof-of-concept implementation can be found [here](https://github.com/TobiasHoll/ctf/tree/master/hxp2020-web-hackme):
 - The main note (`pwn.md`) is responsible for loading the templating JS (and the old jQuery version that it requires) from Disqus, and issuing the two JSONP calls that lead to the DOM injection
 - The `payload.md` note contains the content that is injected into the DOM, and loads and executes the final payload
 - The `win.js` script is the final XSS payload (here, because this was for [hxp 2020 CTF's hackme challenge](https://2020.ctf.link/internal/challenge/d400a07d-ac99-4bfb-ad2a-b6c044541897/) it simply grabs the contents of the `/s/the-flag` page and sends it to the attacker, but you can just as easily run any other JavaScript code).
 - The `pwn.py` script automates the upload process and automatically reports the XSS page (the slides of `pwn.md` in presentation mode) to the challenge admin, which isn't really relevant here.

The most straightforward way to mitigate this is to fix the DOM injection in `reveal-markdown.js` by modifying the check for `</script>` to cover everything that browsers use to end a `<script>` tag. I also wonder what value the `dependency` YAML option really brings to the slides feature - it is restricted to loading JS from pages listed in the CSP anyways, so it would probably be better to just load scripts related to e.g. video embeds as needed (just as in "normal" markdown mode), and discard the option otherwise.

Note that while the CTF challenge was using a (slightly modified, to fix some - but given some of the solutions, apparently not all - dependencies with known CVEs) CodiMD 2.2.0, this exploit still works against the official 2.3.2 release, and [on hackmd.io](https://hackmd.io/p/r1RporhRP) ([source](https://hackmd.io/8MiHR8yMRAeoRDoysn398w)).

jackycute:
Fixes #1648 

Since meta-marked uses js-yaml and parse meta with `yaml.safeLoad`, there is no way the yaml can get js function or regex out. We only need to remove library load external resources.

### DIFF
--- a/public/js/reveal-markdown.js
+++ b/public/js/reveal-markdown.js
@@ -103,7 +103,7 @@ import { md } from './extra'
 
     // prevent script end tags in the content from interfering
     // with parsing
-    content = content.replace(/<\/script>/g, SCRIPT_END_PLACEHOLDER)
+    content = content.replace(/<\/script>/gi, SCRIPT_END_PLACEHOLDER)
 
     return '<script type="text/template">' + content + '</script>'
   }
--- a/public/js/slide.js
+++ b/public/js/slide.js
@@ -80,6 +80,8 @@ const defaultOptions = {
 }
 
 var options = meta.slideOptions || {}
+// delete dependencies to avoid import user defined external resources
+delete options.dependencies
 
 if (Object.hasOwnProperty.call(options, 'spotlight')) {
   defaultOptions.dependencies.push({


jackycute:
Hi @TobiasHoll, could you help us to validate if this fix works?

jackycute:
@TobiasHoll, I've tested the example you attached to #1648.
It seems no longer can reproduce.

PR #1665, opened 2021-03-09
Support anonymous updates via API if allowAnonymousEdits is true

glpatcern:
Hello,

This PR introduces the ability to use the newly provided update API anonymously, when the system is configured to work in anonymous mode (i.e. `allowAnonymousEdits` is true).

Also logging in case of error was extended with the stack trace to match similar code in the project.

The patch was successfully tested and is currently used in a pilot deployment, where CodiMD is being integrated with an external storage and controlled via API, therefore user authentication/authorization is delegated to the external storage. I can provide further information and hopefully more PRs that have helped in our integration.

### DIFF
--- a/lib/note/index.js
+++ b/lib/note/index.js
@@ -292,15 +292,15 @@ const updateNote = async (req, res) => {
         title: Note.parseNoteTitle(content),
         content: content,
         lastchangeAt: now,
-        authorship: req.isAuthenticated() ? [
+        authorship: [
           [
-            req.user.id,
+            req.isAuthenticated() ? req.user.id : null,
             0,
             content.length,
             now,
             now
           ]
-        ] : []
+        ]
       })
 
       if (!updated) {
--- a/lib/note/index.js
+++ b/lib/note/index.js
@@ -268,7 +268,7 @@ const deleteNote = async (req, res) => {
 }
 
 const updateNote = async (req, res) => {
-  if (req.isAuthenticated()) {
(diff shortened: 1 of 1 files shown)


lib/note/index.js
```diff
@@ -292,23 +292,25 @@ const updateNote = async (req, res) => {
         title: Note.parseNoteTitle(content),
         content: content,
         lastchangeAt: now,
-        authorship: [
+        authorship: req.isAuthenticated() ? [
           [
             req.user.id,
```
Yukaii, line 297: We keep authorship for the guest user, set `userId` to `null`:

```suggestion
            req.isAuthenticated() ? req.user.id : null,
```

Yukaii (changes requested):

lib/note/index.js
glpatcern: Hi, makes perfect sense - actually I had thought about updating the `authorship` somehow

Yukaii (approved):

Yukaii:
> I can provide further information and hopefully more PRs that have helped in our integration.

@glpatcern Thanks for the contribution 🎉 . Could you give us more details about the integration?

glpatcern:
> @glpatcern Thanks for the contribution 🎉 . Could you give us more details about the integration?

Sure! I'll tell you more over gitter, the overall approach was simple with no radical changes (e.g. no DB schema changes) but with some UI retouches, following a bit the model of etherpad. @mic4ael has been a great contributor ;-)

PR #1670, opened 2021-04-01
Set Lax for the SameSite attribute on cookies

mic4ael:


mic4ael:
Can anyone have a look at this?

mic4ael:
Actually, this requires a new release of `js-cookie` so changing it to a Draf PR.

jackycute:
Hi @mic4ael,

Thanks for opening this PR.
I think this will take some time to review if case we missed any cookie not been set to Lax.
Also, I'm wondering is there any way we can migrate existing cookie to Lax or the only way is ask user to sign out and sign in again.

PR #1675, opened 2021-04-28
fix: webpack exclude path should support windows path

a60814billy:
1. webpack build will failed in windows


Yukaii:
Thanks @a60814billy

PR #1691, opened 2021-05-31
Strip HTML tags for gist id to avoid stored XSS on showing error [Security Issue]

jackycute:
Fix https://github.com/hackmdio/codimd/issues/1689

cc @lbherrera

Signed-off-by: Max Wu <jackymaxj@gmail.com>

### DIFF
--- a/public/js/extra.js
+++ b/public/js/extra.js
@@ -334,6 +334,14 @@ export function finishView (view) {
       // strip HTML tags to avoid stored XSS
       const gistid = value.getAttribute('data-gist-id')
       value.setAttribute('data-gist-id', stripTags(gistid))
+      const gistfile = value.getAttribute('data-gist-file')
+      if (gistfile) value.setAttribute('data-gist-file', stripTags(gistfile))
+      const gistline = value.getAttribute('data-gist-line')
+      if (gistline) value.setAttribute('data-gist-line', stripTags(gistline))
+      const gisthighlightline = value.getAttribute('data-gist-highlight-line')
+      if (gisthighlightline) value.setAttribute('data-gist-highlight-line', stripTags(gisthighlightline))
+      const gistshowloading = value.getAttribute('data-gist-show-loading')
+      if (gistshowloading) value.setAttribute('data-gist-show-loading', stripTags(gistshowloading))
       $(value).gist(window.viewAjaxCallback)
     }
   })
--- a/public/js/extra.js
+++ b/public/js/extra.js
@@ -330,7 +330,12 @@ export function finishView (view) {
     })
     // gist
   view.find('code[data-gist-id]').each((key, value) => {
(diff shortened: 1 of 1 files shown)


Yukaii (approved):

a60814billy (approved):

jackycute:
> Hi @jackycute,
> 
> I think the fix contained on [85c76f5](https://github.com/hackmdio/codimd/commit/85c76f50a11ed3f5680efff83d11089e065f2d22) is incomplete as the code located in [https://github.com/hackmdio/codimd/blob/develop/public/js/extra.js#L332](https://github.com/hackmdio/codimd/blob/develop/public/js/extra.js?rgh-link-date=2021-06-02T01%3A11%3A08Z#L332) looks for elements that have a code tag with a `data-gist-id` attribute and then sends a reference of it to the gist function of the gist's embed library.
> 
> So if a user created the following note, they would still be able to reach the vulnerable library and get XSS:
> 
> ```
> <code data-gist-id="payload"></code>
> ```
> 
> I think that to fix this issue you will have to make sure that all the attributes of code tags that have a `data-gist-id` attribute are escaped before calling `$(value).gist(window.viewAjaxCallback)`.

Thanks for heads up @lbherrera.
I've force pushed a commit that set attribute before apply gist embed to the element.
Please take a look if that works, thanks.

lbherrera:
I think that works, but there are other attributes that get used by gist embed that also seem to be reflected into error messages and need to be sanitized as well.

```javascript
e = f.data("gist-id") || "";
h = f.data("gist-file");
k = f.data("gist-hide-footer") === true;
d = f.data("gist-hide-line-numbers") === true;
n = f.data("gist-line");
l = f.data("gist-highlight-line");
m = f.data("gist-show-spinner") === true;

[...]

g = f.data("gist-show-loading") !== undefined ? f.data("gist-show-loading") : true
```

I don't see this being exploitable anymore if you also sanitize the `data-gist-file`, `data-gist-line`, `data-gist-show-loading` and `data-gist-highlight-line` attributes.

jackycute:
> I think that works, but there are other attributes that get used by gist embed that also seem to be reflected into error messages and need to be sanitized as well.
> 
> ```js
> e = f.data("gist-id") || "";
> h = f.data("gist-file");
> k = f.data("gist-hide-footer") === true;
> d = f.data("gist-hide-line-numbers") === true;
> n = f.data("gist-line");
> l = f.data("gist-highlight-line");
> m = f.data("gist-show-spinner") === true;
> 
> [...]
> 
> g = f.data("gist-show-loading") !== undefined ? f.data("gist-show-loading") : true
> ```
> 
> I don't see this being exploitable anymore if you also sanitize the `data-gist-file`, `data-gist-line`, `data-gist-show-loading` and `data-gist-highlight-line` attributes.

Thanks for the reminder.
I've pushed commit that also sanitize those attributes, please take a look.

PR #1697, opened 2021-06-12
fix: should not clear guest history when guest pin note

a60814billy:
as title


Yukaii:
Force push to retrigger the action build

jackycute:
Thanks @a60814billy and @Yukaii

Issue #1478, opened 2020-04-09
Feature Request: ability to choose custom URL for a note after creation


crosswick:
Being able to choose a free URL by simply typing it is nice - what would also be nice is if you could do a 'save as' so to speak: choose a free URL after the note has been created.

Might also be nice to have a dialog box open after you click New Note, asking for a custom URL, with the option of "don't ask this again".

And/or: a text entry field next to the New Note button for entering a custom URL before clicking, that checks for availability right away.

stratosgear:
How do you "choose a free URL by simply typing it"?

You mean you visit the: `my.domain.com/someUrl` and create a note...? Because I cannot get this to work!

PR #1699, opened 2021-06-19
Feat/custom note url

Nick0603:
Relatied issue: [#1478](https://github.com/hackmdio/codimd/issues/1478)

story:

1. A created note
image

2. Click "Custom Note Url" in Menu
image

3. Type url you like and submit.
image

4. The url is customized and sync to all collaborators in real time.
image

Details:

1. That's ok if you want to set custom url for many times.
- /serial_number -> /custom1 -> /custom2 -> /custom3

If some users browse the past url, they will redirect to latest one automatically.
- /serial_number -> /custom3
- /custom1 -> /custom3
- /custom2 -> /custom3

2. It will be blocked when the custom url is conflict with exist url.
image

3. But if the conflicted url is set by you in the past. You have right to set back.
-  /custom1 -> /custom2 -> /custom1

4. Only the owner of note can edit custom url.
image


Nick0603:
Fixed the linter error.

lib/note/service.js
```diff
@@ -0,0 +1,108 @@
+
+const { Note, ArchivedNoteAlias, sequelize } = require('../models')
+const realtime = require('../realtime/realtime')
+
+const forbiddenAlias = ['', 'new', 'me', 'history', '403', '404', '500', 'config']
+const sanitize = (alias) => {
+  return alias.replace(/( |\/)/, '')
```
Yukaii, line 7:
```suggestion
  return alias.replace(/( |\/)/g, '')
```

Yukaii (changes requested):

lib/note/service.js
```diff
@@ -0,0 +1,108 @@
+
+const { Note, ArchivedNoteAlias, sequelize } = require('../models')
+const realtime = require('../realtime/realtime')
+
+const forbiddenAlias = ['', 'new', 'me', 'history', '403', '404', '500', 'config']
```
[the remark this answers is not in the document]
jackycute: Yes

lib/models/note.js
```diff
@@ -223,6 +228,18 @@ module.exports = function (sequelize, DataTypes) {
 
   Note.parseNoteId = function (noteId, callback) {
     async.series({
+      parseNoteIdByArchivedAlias: function (_callback) {
```
jackycute, line 231: I'm not totally sure about why we keep the previously used alias in track?
What if the archived alias you about to change has been used?

a60814billy (changes requested):
Hi @Nick0603,

Thanks for your contribution, there are some suggestions:
1. In general, we use strict mode in backend. 
2. Do not mix different asynchronous paradigm in a procedure.
3. Maybe we can fill-in alias url in modal if the note already setup an alias url.
4. Redirect to alias url when user navigate via noted
5. To avoid some vulnerability, we only support these character in customize url: lowercase
   letters, decimal digits, hyphen, underscore. And need check user input is valid both frontend and backend. Do not auto correct or filter out the input.

public/js/index.js
```diff
@@ -2217,6 +2285,11 @@ socket.on('cursor blur', function (data) {
   }
 })
 
+socket.on('alias updated', function (data) {
+  const alias = data.alias
+  history.replaceState({}, '', alias)
```
a60814billy, line 2290: Why do you need this line?

lib/note/service.js
```diff
@@ -0,0 +1,108 @@
...
+const sanitize = (alias) => {
+  return alias.replace(/( |\/)/, '')
+}
+
+const asyncGetNote = async (originAliasOrNoteId) => {
```
a60814billy, line 10: We already have `getNoteById` function in note/index.js

jackycute:
I think we need some redirection from `/:noteId` to `/:alias` URL to help user who saved noteId URL to browse note with new alias URL.

lib/note/service.js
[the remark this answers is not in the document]
a60814billy: Also config.forbiddenNoteIDs

lib/note/index.js
```diff
@@ -8,6 +8,7 @@ const { newCheckViewPermission, errorForbidden, responseCodiMD, errorNotFound, e
 const { updateHistory, historyDelete } = require('../history')
 const { actionPublish, actionSlide, actionInfo, actionDownload, actionPDF, actionGist, actionRevision, actionPandoc } = require('./noteActions')
 const realtime = require('../realtime/realtime')
+const serv = require('./service')
```
a60814billy, line 11: rename service -> noteAlias, and consider using object destruction

lib/note/service.js
```diff
@@ -0,0 +1,108 @@
...
+      alias: updatedNote.alias
+    })
+
+  return true
+}
+
+exports.sanitize = sanitize
+exports.asyncGetNote = asyncGetNote
+exports.asyncCheckAliasValid = asyncCheckAliasValid
+exports.asyncUpdateAlias = asyncUpdateAlias
```
a60814billy, line 108: Put async in postfix not prefix.
e.g. asyncUpdateAlias -> updateAliasAsync

lib/migrations/202106200000000-create-note-aliases-history.js
```diff
@@ -0,0 +1,32 @@
...
+        () => queryInterface.addIndex(
+          'ArchivedNoteAliases',
+          ['alias'], {
+            indicesType: 'UNIQUE'
+          })
+      )
+  },
+
+  down: function (queryInterface) {
+    return queryInterface.dropTable('Users')
```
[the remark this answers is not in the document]
Nick0603: Sorry for typo

lib/models/note.js
Nick0603: > I'm not totally sure about why we keep the previously used alias in track?
My thought is to avoid some user saved the archived alias but can't connect the note. 

> What if the archived alias you about to change has been used?
The user just change one that not be used. Or we need to set the special and personal prefix for the alias?

jackycute: To make this PR specific, it's better not to add this feature for redirecting previous alias to new one.
I suggest to remove this and just adding a notice alert area in the custom note url modal like: `Notice: the previous custom note url will not be redirected to the new one, please take your own risk to change this.`

lib/note/service.js
Nick0603: Because the `getNoteById` in note/index.js noget pure note row data.
I think `controller` or `service` should be choose the enough data to use.

lib/note/index.js
Nick0603: I would like to make the service is for all note controller.
And the service will contains the common logic in controller and easy for testing.

Nick0603:
> Hi @Nick0603,
> 
> Thanks for your contribution, there are some suggestions:
> 
> 1. In general, we use strict mode in backend.
> 2. Do not mix different asynchronous paradigm in a procedure.
> 3. Maybe we can fill-in alias url in modal if the note already setup an alias url.
> 4. To avoid some vulnerability, we only support these character in customize url: lowercase
>    letters, decimal digits, hyphen, underscore. And need check user input is valid both frontend and backend. Do not auto correct or filter out the input.
> 5. Alias would be conflict with another nodeId, shortid


1. OK, I will add the missed`'use strict'` line in code
2. I don't really know the suggestion. Could you say more or give me a example.
3. Yes, now the code can do it.
4. OK
5. OK, I use `parseNoteIdAsync` to validate the alias inputed by user. And then alias will be check by all aliases for a noteId includes 'note id '、'note short id'、'public/docs/*.md'

a60814billy:
> Do not mix different asynchronous paradigm in a procedure.

For example: 

The code mix two asynchronous paradigm, async/await and Promise.
```js
const conflictNote = await exports.getNote(alias)
  .catch((err) => { throw err })
```

If you want to handle error in async/await, using try...catch statement.

```js
try {
  const conflictNote = await exports.getNote(alias)
} catch ((err) => {
  throw err
})
```

But, the catch statement just rethrow the error, we can remove it.

```js
const conflictNote = await exports.getNote(alias)
```

lib/note/index.js
```diff
@@ -332,9 +253,43 @@ const updateNote = async (req, res) => {
...
+  const userId = req.user ? req.user.id : null
+  const note = await service.getNote(originAliasOrNoteId)
+    .catch((err) => {
+      logger.error('get note failed:' + err.message)
+      return false
+    })
+
+  if (!note) {
+    logger.error('update note alias failed: note not found.')
+    return res.status(500).json({ status: 'error', message: 'Internal Error' })
```
a60814billy, line 268: return 404 for note not found

public/js/index.js
```diff
@@ -1266,6 +1277,62 @@ $('#revisionModalRevert').click(function () {
   editor.setValue(revision.content)
   ui.modal.revision.modal('hide')
 })
+
+// custom note url modal
+const updateNoteUrl = (noteUrl = '') => {
```
a60814billy, line 1282: rename: `const updateNoteAliasUrl = (noteAliasUrl = '')`
or `const updateNoteCustomUrl = (noteCustomUrl = '')`

public/js/index.js
```diff
@@ -1266,6 +1277,62 @@ $('#revisionModalRevert').click(function () {
...
+  }
+  const hideErrorMessage = () => ui.modal.customNoteUrl.find('.js-error-alert').hide()
+
+  const customUrl = ui.modal.customNoteUrl.find('[name="custom-url"]').val()
+  if (!/^[0-9a-z-_]+$/.test(customUrl)) {
+    showErrorMessage('The url must be lowercase letters, decimal digits, hyphen or underscore.')
+    return
+  }
+
+  updateNoteUrl(customUrl)
```
a60814billy, line 1311: handle error in catch statement:

```js
updateNoteUrl(customUrl)
  .then(({status}) => { .... })
  .catch((err) => {
    if (err.status === 400) .....
    ...
    showErrorMessage('Something wrong.')
  })
```

a60814billy (changes requested):

a60814billy:
Big thanks @Nick0603

PR #1707, opened 2021-08-12
WIP: add password permission

MROS:


jackycute:
Hi @MROS,

Thanks for opening this PR.
However, we think this implementation was a bit different from what's described in the issue, the note password should validate with permission instead of adding another permission for password.
I will close this PR and feel free to open another one if you like to rework, thank you!

PR #1710, opened 2021-08-25
feat: add organizations whitelist to GitHub OAuth

jakubgs:
Currently CodiMD does not support limiting access of GitHub OAuth users based on their organization membership. This is a very useful functionality for teams that want to limit write access to their notes.

I've implemented a crude solution to this problem, which most probably requires some adjusments to make it better. I'm not sure if this implementation is kosher, but it definitely works on my deployment.

Open to suggestions on how I can improve it.

### DIFF
--- a/app.json
+++ b/app.json
@@ -80,6 +80,14 @@
             "description": "GitHub API client secret",
             "required": false
         },
+        "CMD_GITHUB_ORGANIZATIONS": {
+            "description": "GitHub whitelist of orgs",
+            "required": false
+        },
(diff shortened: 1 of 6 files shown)


nderjung:
Hey this is a pretty useful feature, any update @a60814billy, @jackycute?

jackycute:
Hi @jakubgs @nderjung,

We will put this into the next release.

jackycute:
That's try to merge this.
Thank you @jakubgs

PR #1716, opened 2021-09-27
Upgrade codemirror to 5.61.3

Yukaii:


jackycute:
Maybe we can upgrade to 5.63.2?

PR #1739, opened 2022-01-06
Fix: Code Fence parameter parsing

### LINKED ISSUE #1738: Parsing Code Fence Parameters causes TypeError
## Context

I'm using code fence parameters for my pandoc pipeline to render plantuml diagrams using pandoc's diagram filter. Unfortunately, the mandatory syntax ` ```{#fig:use-cases .plantuml caption="foo"} ` causes codimd to throw an error while parsing the parameters.

## Workaround

Switch to Edit-Mode (by changing the URL's query param to `?edit`) and remove the causing line.

## Cause

codimd is capable of parsing the expression above, the function is called `parseFenceCodeParams`. There is a Not-Operator missing that causes the `class` property to never be initialized. The next line fails due to the uninitialized property.

https://github.com/hackmdio/codimd/blob/ded6b5856f988ec6900a341a33a164a0c1ae3e12/public/js/lib/markdown/utils.js#L12-L13

V1ncNet:
The change fixes an error while rendering diagrams using code fence parameters.

This closes #1738

### DIFF
--- a/public/js/lib/markdown/utils.js
+++ b/public/js/lib/markdown/utils.js
@@ -9,7 +9,7 @@ export function parseFenceCodeParams (lang) {
       if (param[0] === '#') {
         params.id = param.slice(1)
       } else if (param[0] === '.') {
-        if (params.class) params.class = []
+        if (!params.class) params.class = []
         params.class = params.class.concat(param.slice(1))
       } else {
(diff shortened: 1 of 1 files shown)


jackycute:
Thanks @V1ncNet

PR #1741, opened 2022-01-16
Update de.json

rubstudent:
better german

PR #1701, opened 2021-07-05
Add arm64 docker image build.

YadominJinta:
Need to use the runtime and buildpack image in [hackmdio/docker-buildpack#3](https://github.com/hackmdio/docker-buildpack/pull/3)

I really hope that you can publish an arm64 build in dockerhub.

PR #1688, opened 2021-05-25
Add dev container for GitHub Codespaces and VSCode remote container

a60814billy:
This PR add dev container support.

refs:
- https://docs.github.com/en/codespaces/setting-up-your-codespace/configuring-codespaces-for-your-project
- https://code.visualstudio.com/docs/remote/containers

PR #1683, opened 2021-05-13
Fix S3 client config passing for image upload

jackycute:
Fix #1572 which not passing config to S3 client properly

ref issue: https://github.com/aws/aws-sdk-js-v3/issues/1396

Signed-off-by: Max Wu <jackymaxj@gmail.com>

PR #1742, opened 2022-01-18
Release 2.4.2

a60814billy:
## Security Fixes
- #1685 **[Security Issue]** Upgrade flowchart.js to version 1.15.0
- #1690 **[Security Issue]** Upgrade mermaid to 8.10.2
- #1691 **[Security Issue]** Strip HTML tags for gist id to avoid stored XSS on showing error
- #1695 **[Security Issue]** Add CSRF token in export API to prevent security issue
- #1716 **[Security Issue]** Upgrade CodeMirror to 5.63.2

## Fixes

- #1605 Fix container syntax not parsed correctly
- #1679 Handle when request url has no valid referer
- #1683 Fix S3 client config passing for image upload
- #1706 Fix array access index may out of bound
- #1723 remove unused uglifyjs webpack plugin dependency

## Enhancements

- #1481 Set lang attributes via user locale
- #1580 Use include function instead of directives
- #1587 Extract more keyword for i18n translate
- #1680 Avoid append zero suffix on exporting user data
- #1684 Add TeX mhchem extensions for MathJax
- #1701 Support arm64 docker image
- #1724 Refactor Sequelize model import mechanism due to sequelize.import is deprecated
- #1741 Better german translation

## DX

- #1688 Support DevContainer for GitHub Codespaces and VSCode remote container
- #1694 Run CI with GitHub Actions


jackycute (approved):

jackycute (changes requested):
I think this PR should be version 2.4.2 instead?

productiveme:
Wondering if this will merge and release any time soon?

inductor:
@a60814billy @jackycute What's the status of this PR?

jackycute (approved):
Thanks @a60814billy

PR #1750, opened 2022-03-03
🐛 [fix] modify replacement rule for disqus short-name

chenxuanzzy:
Modify replacement rule for disqus short-name.


jackycute:
This looks good to me, thanks @chenxuanzzy

PR #1767, opened 2022-08-02
Upgrade Node.js version

inductor:


a60814billy (approved):
Thanks @inductor, we can move forward to Node 14 in next version!

jackycute:
Thanks @inductor @a60814billy

PR #1790, opened 2023-01-07
FIX: pandoc security issue

galaxian85:
According to pandoc's [manual](https://pandoc.org/MANUAL.html#a-note-on-security)
some file format can include or embed files on the file system.

Check param `exportType` is valid & running pandoc with `--sandbox` to fix this.


jackycute:
Thanks @galaxian85

PR #1792, opened 2023-01-10
potential XSS in vimeo embed

galaxian85:
access `videoid` directly has potential XSS risk


jackycute:
Thanks @galaxian85,
This should fix https://github.com/hackmdio/codimd/issues/1774

jackycute:
Thanks @galaxian85

PR #1797, opened 2023-02-14
fix(buildpacks): replace custom buildpack with APT buildpack

EtienneM:
Fix the Scalingo deployment by using the Scalingo official recommandation for VIPS (https://doc.scalingo.com/languages/ruby/rails/start#use-vips-to-process-image).


jackycute (approved):

jackycute:
Thanks @EtienneM

PR #1799, opened 2023-03-01
Update minimum required node.js version to v12 with npm package dependencies

PeterDaveHello:
Looks like the node.js and npm dependencies didn't get update for a while, node.js v10 is EoL at 2021-04-30, and there could be a lot of security vulnerabilities since the last release, especially for those using pre-built Docker image on Docker Hub.

The idea here, is try to increase the baseline of the node.js and npm packages "a little bit", and ideally will update to node.js v18, because both of node.js v14 & v16 will reach EoL this year, node.js v12 is EoL at 2022-04-30 last year.

As I can't guarantee how much I can contribute, each PR won't be large, so that we can get it reviewed and merged in a more agile way ;) 

Currently, `npm run test` is the assurance of the update not breaking things, please let me know if that's not enough. Thanks.

Reference:

- Prompt message from `npm install` before the update:

    > found 350 vulnerabilities (13 low, 163 moderate, 138 high, 36 critical)

- Prompt message from `npm install` after the update:

    > found 123 vulnerabilities (13 low, 42 moderate, 52 high, 16 critical)


PeterDaveHello:
Looks like there are some thing need to be fixed.

package.json
```diff
@@ -164,9 +164,9 @@
     "markdown-it-ruby": "^0.1.1",
     "markdown-it-sub": "~1.0.0",
     "markdown-it-sup": "~1.0.0",
-    "markdownlint": "^0.22.0",
+    "markdownlint": "^0.27.0",
     "markdownlint-rule-helpers": "^0.13.0",
-    "markmap-lib": "^0.4.2",
+    "markmap-lib": "^0.14.4",
```
PeterDaveHello, line 169: Update to v0.14 seems to fast that needs additional changes, will downgrade to v0.6 so the build won't fail.

jackycute (approved):

PR #1802, opened 2023-03-23
【fix】reword japanese

AQ-masatoshi-yamaguchi:
Hi!
Corrected as Kanji is preferable to Hiragana.

Thank you.


jackycute:
Thanks @AQ-masatoshi-yamaguchi

PR #1813, opened 2023-08-09
Add the logout callback to prevent exception.

hcyuser:


hcyuser (commented):
add callback for logout

jackycute:
Hi @hcyuser,

Thanks for contribution.
Can you explain what is this PR solving?

hcyuser:
If we don't add a callback in req.logout(callback),
It will get _Error: req#logout requires a callback function_ on the webpage when running node app.JS and clicking Sign out on the webpage.

jackycute:
Thanks for adding this.
This is caused by https://stackoverflow.com/questions/72336177/error-reqlogout-requires-a-callback-function from https://github.com/jaredhanson/passport/issues/902#issuecomment-1133897569

lib/auth/index.js
```diff
@@ -53,6 +53,8 @@ authRouter.get('/logout', function (req, res) {
   if (config.debug && req.isAuthenticated()) {
     logger.debug('user logout: ' + req.user.id)
   }
-  req.logout()
+  req.logout(()=>{
+    logger.debug('Please implement logout callback here.')
+  })
```
jackycute, line 58: It should be like:

```suggestion
  req.logout((err) => {
    if (err) { return next(err) }
    res.redirect(config.serverURL + '/')
  })
```

and add `next` as the last parameter for get route callback function

brent85 (approved):

lib/auth/index.js
jackycute: Hi @hcyuser,
Please consider this suggestion.

Yukaii:
closed in favor of #1837

PR #1848, opened 2024-01-04
chore/build image with github action

Yukaii:
- chore: remove cjk image build
- chore: push-image workflow

### DIFF
--- a/.github/workflows/push-image.yml
+++ b/.github/workflows/push-image.yml
@@ -1,0 +1,123 @@
+name: Build and push image
+
+on:
+  release:
+    types: [published]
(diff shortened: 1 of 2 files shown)


.github/workflows/push-image.yml
```diff
@@ -0,0 +1,123 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - linux/amd64
+          # - linux/arm64
```
Yukaii, line 28: Disabled for now. Will come back to enable it after everything works fine

Issue #1601, opened 2020-09-19
Unexpected behaviour: Uploaded images are accessible for everyone


RayMacLewis:
When I upload an image (stored on the server itself) the image is not protected like the *.md-file is.
I understand the problem to keep another access-level for each image but if I mark a note as private I also expected the pictures to be private only. In fact I can even access them without being logged in just by having the URL.

PR #1866, opened 2024-05-01
[WIP] Move image path to the session scope.

shohhei1126:
Work in progress for #1601


jackycute:
Hi @shohhei1126,

Thanks for opening PR!

There is a use case that user upload images in the note and later they like to use those image urls on the other platform (like external blog post or SNS) and those images will become inaccessible due to the session auth.
This is the main reason that the uploaded images are public accessible by current design.

brent85 (approved):

PR #1867, opened 2024-05-27
refactor: reuse random filename in filesystem image provider

Yukaii:


lib/imageRouter/filesystem.js
```diff
@@ -47,7 +29,7 @@ exports.uploadImage = function (imagePath, callback) {
 
   let filename = path.basename(imagePath)
   try {
-    filename = pickFilename(path.basename(imagePath))
+    filename = randomFilename()
```
jackycute, line 32: the filename will no longer with file extension (.png), we should modify pickFilename function and only use randomFilename their without default filename.

lib/imageRouter/filesystem.js
```diff
@@ -22,7 +22,7 @@ function randomFilename () {
  */
 function pickFilename (defaultFilename) {
   let retryCounter = 5
-  let filename = defaultFilename
+  let filename = randomFilename()
   const extname = path.extname(defaultFilename)
   while (retryCounter-- > 0) {
     if (fs.existsSync(path.join(config.uploadsPath, filename))) {
```
jackycute, line 28:
```suggestion
  const extname = path.extname(defaultFilename)
  let filename = `${randomFilename()}${extname}`
  while (retryCounter-- > 0) {
    if (fs.existsSync(path.join(config.uploadsPath, filename))) {
```

jackycute (approved):

PR #1879, opened 2024-08-08
Fix issue with SQlite connection

alibell:
The current docker image doesn't work with SQLite.

There are two main issues :
- pcheck is executed even if the $CMD_DB_URL connection is configured to use an sqlite database
- the sqlite3 node package is not installed by default

The current fix resolve these two issues.


Nexyll:
I was wondering why a migration was created to add an index in this PR?
By the way, I’m very interested in the fix. Is the project still active?

jackycute:
Hi @alibell,
Thanks for the PR.
However, according to the description of the PR, I think the Revision migration script should be separated to another PR.

PR #1882, opened 2024-09-24
event.js updated for dark mode and puppet.js introduced for download html function

chetanpilane:


jackycute:
Hi @chetanpilane,
Thanks for the PR.
But I don't quite understand what is this PR for, can you elaborate?

Issue #1892, opened 2025-02-20
Lost schema of S3 URL.


fukumoto-ts:
Since v2.5.0, S3 URL schema is now lost.

<img width="467" alt="Image" src="https://github.com/user-attachments/assets/bb03965c-0d39-4a52-a1f9-e752ba42dc64" />

This is caused by a change in this PR. #1762 (#1761)

---
It would be good to change it like this.

Now:
https://github.com/blademainer/codimd/blob/7b1e4e3ddb2ef07d1170dcd819986252d5758596/lib/imageRouter/s3.js#L54

```
s3Endpoint = `s3-${config.s3.region}.amazonaws.com`
```

After:

```
s3Endpoint = `https://s3-${config.s3.region}.amazonaws.com`
```

PR #1893, opened 2025-02-20
Fix: Lost schema of S3 URL. 

fukumoto-ts:
close: #1892

Fix the erroneous change in #1762.


jackycute (approved):
LGTM, thanks @fukumoto-ts

PR #1894, opened 2025-03-19
Upgrade mermaid to latest version

Yukaii:
Mark extraneous dependencies in package-lock.json as dev and optional, and update the mermaid dependency to version 11.5.0.


jackycute:
Hmm, the CI build test failed.

Yukaii:
It seems that node 14.x is no longer supported with the current dependencies. I'll just remove it now. (Before we upgrade to stable node 20.x, 16.x should also be deprecated)

PR #1899, opened 2025-03-21
feature/dev 1879 improve contrast on certain elements

stanley2058:
Improve contrast of UI elements:
- Default theme now has higher contrast linenumber & code comments.
The link buttons in the navbar, last changed at, and the action button on the top right are now more visible in dark mode.
- Fix an issue where the theme name would not be visible when hovering under dark mode.

Before:
Screenshot_20250321_115308

After:
Screenshot_20250321_115243

The color of active line is also updated:
(Before)
image
(After)
image


jackycute (approved):
It looks much better now!

PR #1908, opened 2025-04-14
chore/bump codemirror version to 5.65.8

stanley2058:
Bumping Codemirror to version 5.65.8 addressed an issue on Firefox: The editor would break if the user tried to input after right-clicking on it.


jackycute (approved):
LGTM!

PR #1911, opened 2025-04-24
fix: login email should be case insensitive

### LINKED ISSUE #1744: email lowercase
when checking for the password the email should be put through tolowercase before comparing because no one remembers when one used upper and lowercase and as it doesnt matter in emails what you spell upper and lower case; most services ignore that already and so should codimd.

DubMythlit:
fix #1744
By query email in lower case makes login be case insensitive now
tested on: Postgres12 & sqlite3

### DIFF
--- a/lib/auth/email/index.js
+++ b/lib/auth/email/index.js
@@ -2,6 +2,7 @@
 
 const Router = require('express').Router
 const passport = require('passport')
+const sequelize = require('sequelize')
 const validator = require('validator')
 const LocalStrategy = require('passport-local').Strategy
 const config = require('../../config')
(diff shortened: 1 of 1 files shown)


lib/auth/email/index.js
```diff
@@ -21,7 +22,11 @@ passport.use(new LocalStrategy({
   try {
     const user = await models.User.findOne({
       where: {
-        email: email
+        email: sequelize.where(
+          sequelize.fn('LOWER', sequelize.col('email')),
+          'LIKE',
```
jackycute, line 27: I don't think we need to use LIKE operator here, compare directly will be fine.

jackycute:
Hi @Dubmyhtlit,

Thanks for your PR,
Please also sign off your commit like this:

```
commit message

Signed-off-by: XXX <xxx@example.com>
```

Yukaii:
@DubMythlit You have to signoff every commit:

```bash
git rebase --signoff -i develop
```

Yukaii (approved):

PR #1912, opened 2025-05-08
Fix login redirection to requested page after authentication

Nowaker:
# Problem

When a user attempts to access a protected resource without being logged in, they are correctly redirected to the login page with a `next` parameter preserving their intended destination (e.g., `/?next=%2Fwb0e9nB5T3qecF27Y4xlsg`). However, after successful authentication, the application always redirects to the dashboard/homepage instead of returning the user to their originally requested URL.

This creates a poor user experience as users must manually navigate back to the content they were originally trying to access.

# Solution

The root cause was identified in the email authentication implementation where the `successReturnToOrRedirect` parameter was hardcoded to always redirect to the server's homepage.

```javascript
successReturnToOrRedirect: config.serverURL + '/',
```

This change properly utilizes the `req.session.returnTo` value that's already being correctly set by the `setReturnToFromReferer` function:

```javascript
// After: Redirect to originally requested page or homepage if none
successReturnToOrRedirect: req.session.returnTo,
```

# Testing
Manually tested the authentication flow by:

1. Attempting to access a protected note while logged out
2. Confirming redirection to login page with proper next parameter
3. Logging in with valid credentials
4. Verifying successful redirection to the originally requested note

This fix respects the standard behavior of Passport.js's `successReturnToOrRedirect` parameter, which will redirect to the URL stored in `req.session.returnTo` if available, or fall back to the configured redirect URL otherwise.


jackycute:
Hi @Nowaker,

Thank you for your PR!
Please also sign off you commit and force push again :)
```
git rebase --signoff -i develop
```

P.S. Glad to see you here, howdy!

Nowaker:
@jackycute Howdy howdy! :)

jackycute (approved):

Yukaii (approved):

Yukaii (changes requested):
@Nowaker Could you also update other authentication provider, like

https://github.com/hackmdio/codimd/blob/b4ec353fcd5cfdee333373087ea9ff386f2fbea7/lib/auth/facebook/index.js#L26

PR #1940, opened 2025-09-16
docs: edited broken badge travis

bravesasha:


Yukaii:
@bravesasha You can use

```bash
git commit --amend --signoff
```

And do a force push

